PR #17, opened 2021-12-21
data models and validation

nebs-dev:


src/entities/Period.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export interface PeriodDocument extends PeriodInput, mongoose.Document {
+  createdAt: Date;
+  updatedAt: Date;
+}
+
+const periodSchema = new mongoose.Schema(
+  {
+    id: { type: String, required: true, unique: true },
```
kristoferlund, line 17: Doesn't Mongoose create a `_id` for you automatically? If so we don't need this id.

src/entities/Praise.ts
```diff
@@ -0,0 +1,59 @@
...
+  },
+  {
+    timestamps: true,
+  }
+);
+
+const praiseSchema = new mongoose.Schema(
+  {
+    id: { type: String, required: true, unique: true },
+    periodId: { type: String, required: true },
```
kristoferlund, line 46: Shouldn't we reference period using `ObjectId`?

```
{ type: Schema.Types.ObjectId, ref: 'Period' }
```

And, since periods are created manually, it can't be `required: true`.

kristoferlund: https://mongoosejs.com/docs/populate.html

src/entities/Praise.ts
```diff
@@ -0,0 +1,59 @@
...
+    reason: { type: String, required: true },
+    quantifications: [QuantificationSchema],
+    giver: AccountSchema,
```
kristoferlund, line 49: Can giver and receiver be `required: true`?

src/entities/Praise.ts
```diff
@@ -0,0 +1,59 @@
...
+    mongoose.Document {
+  createdAt: Date;
+  updatedAt: Date;
+}
+
+const QuantificationSchema = new mongoose.Schema(
+  {
+    id: { type: String, required: true, unique: true },
+    quantifierId: { type: String, required: true },
+    score: { type: Number, required: true },
```
kristoferlund, line 34: Score is not required. When we assign quantifiers we create quantification objects without scores.

src/entities/Praise.ts
```diff
@@ -0,0 +1,59 @@
...
+    dismissed: { type: Boolean, required: true, default: false },
+    duplicatePraiseId: { type: String },
```
kristoferlund, line 36: Reference `ObjectId` instead?

PR #21, opened 2021-12-27
A few period endpoints

kristoferlund:


src/controllers/periods.ts
```diff
@@ -1,18 +1,63 @@
...
+  return res.status(StatusCodes.OK).json(response);
 };
 
-export default { getPeriods };
+export const single = async (
+  req: Request<any, QueryInput, any>,
+  res: Response
+): Promise<Response> => {
+  let period = await PeriodModel.findById(req.params.periodId);
+  if (!period) return res.status(StatusCodes.INTERNAL_SERVER_ERROR);
```
nebs-dev, line 23: it is a small thing but why are we using `let` instead of `const` here? and maybe we should return `NOT_FOUND` in cases like this?

src/controllers/periods.ts
```diff
@@ -1,18 +1,63 @@
...
+};
+
+export const update = async (
+  req: Request<any, QueryInput, any>,
+  res: Response
+): Promise<Response> => {
+  const name = req.query.name as string;
+  const endDate = req.query.endDate as string;
+
+  if (!name && !endDate) return res.status(StatusCodes.BAD_REQUEST);
```
nebs-dev, line 34: i created `QueryInput` for the query params that will be used for list endpoint, for exaple limit, page, etc :
```export interface QueryInput extends Query {
  sortColumn?: string;
  sortType?: string;
  limit?: string;
  page?: string;
}```

i though that on update params will go to the body of the request. if so we can create periodUpdateParams interface inside `inputs.ts` and than request could look like this:
```Request<RequestBody, any, any>```

That way we know exactly how request will look like and we don't need to check if `name` and `endDate` exist inside the request

File: src/controllers/periods.ts (hunk 1 above)
Comment (kristoferlund) [reply src/controllers/periods.ts:23]: No reason. Will fix.

File: src/controllers/periods.ts (hunk 2 above)
Comment (kristoferlund) [reply src/controllers/periods.ts:34]: Ok, all update params in request body.

We still need to check object received though since TypeScript does not do exact types. 

https://github.com/microsoft/TypeScript/issues/12936

If interface is: 

```
interface periodUpdateParams {
  name?: string;
  endDate?: string;
}
```

The following object would not generate an error:

```
{
  name: "new name",
  paramWeDontWantUserToChange: "new value"
}
```

PR #39, opened 2022-01-13
Import praise script

nebs-dev:


kristoferlund:
Import script can be run with yarn: 
```
yarn import-praise file.json
```

PR #47, opened 2022-01-19
Feat/discord-bot/praise command

Vyvy-vi:
## Description

 - This refactors the command Registering process a bit.
 - This PR adds the `/praise` command.
 
`/praise` command:
- Syntax `praise receivers: string reason: string`
- The command parses out the receiving users from the `receivers` variable using this regex - `receivers?.match(/<@!([0-9]+)>/g)`
- This command might be slightly expensive for cases with more than 20-30 receivers.

This updates the `.env.template` file, and expects a few more variables to exist in there-
- `DISCORD_TOKEN` (the discord bot token)
- `DISCORD_CLIENT_ID` (the client ID for the bot application client)
- `DISCORD_GUILD_ID` (the snowflake ID of the guild in which this bot is supposed to process incoming Praise)
- `PRAISE_GIVER_ROLE_ID` (the snowflake ID of the role which would be able to give praise) (this can be turned into a setting on the frontend in a later iteration)

Another update:
- This requires the `GUILD_MEMBERS` Intent permissions. This requires the bot to have that enabled, since that's  

## Demo

https://user-images.githubusercontent.com/62864373/150232363-a32f6fcc-9cf0-4eaa-be44-3a11adbf1bc2.mp4

(in above, you might notice that the last case doesn't offer a warning to the user. This can be cleared up in a second PR, when refactoring this code)


src/bots/discord/commands/praise.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  const praiseGiverRole = guild.roles.cache.find(
+    (r) => r.id === process.env.PRAISE_GIVER_ROLE_ID
+  );
+  const praiseGiver = await guild.members.fetch(member.user.id);
+
+  if (!praiseGiver.roles.cache.find((r) => r.id === praiseGiverRole?.id)) {
+    const noGiverRoleEmbed = praiseErrorEmbed(
+      `User does not have \`${praiseGiverRole?.name}\` role`,
+      `The praise command can only be used by members with the <@&${praiseGiverRole?.id}> role. Attend an onboarding-call, or ask a steward or guide for an Intro to Praise.`
```
kristoferlund, line 31: Let's use the settings system for this message.

src/bots/discord/commands/praise.ts
```diff
@@ -0,0 +1,199 @@
...
+    await interaction.reply({ embeds: [noGiverRoleEmbed] });
+    return;
+  }
+
+  const ua = {
+    id: member.user.id,
+    username: member.user.username + '#' + member.user.discriminator,
+    profileImageUrl: member.user.avatar,
+    platform: 'DISCORD',
+    activateToken: randomstring.generate(),
```
kristoferlund, line 42: No need to generate this here. It is done when `/praise-activate` is called

src/bots/discord/commands/praise.ts
```diff
@@ -0,0 +1,199 @@
...
+  };
+
+  const userAccount = await UserAccountModel.findOneAndUpdate(
+    { username: ua.username },
```
kristoferlund, line 46: Find user using `id` instead.

src/bots/discord/commands/praise.ts
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  const User = await UserModel.findOne({
+    accounts: userAccount,
+  });
+
+  if (!User) {
+    const notActivatedEmbed = praiseErrorEmbed(
+      'Account Not Activated',
+      'Your Account is not activated in the praise system. Unactivated accounts can not praise users. Use the `/praise-activate` command to activate your praise account and to link your eth address.'
```
kristoferlund, line 114: Let admins configure this using settings.

src/bots/discord/commands/praise.ts
```diff
@@ -0,0 +1,199 @@
...
+    interaction?.channel?.id || ''
+  );
+
+  for (const receiver of Receivers) {
+    const ra = {
+      id: receiver.user.id,
+      username: receiver.user.username + '#' + receiver.user.discriminator,
+      profileImageUrl: receiver.avatar,
+      platform: 'DISCORD',
+      activateToken: randomstring.generate(),
```
kristoferlund, line 138: Same as above, no need here.

src/bots/discord/commands/praise.ts
```diff
@@ -0,0 +1,199 @@
...
+    };
+    const receiverAccount = await UserAccountModel.findOneAndUpdate(
+      { username: ra.username },
```
kristoferlund, line 141: Find on `id`

src/bots/discord/commands/praise.ts
```diff
@@ -0,0 +1,199 @@
...
+      ra,
+      { upsert: true, new: true }
+    );
+
+    const receiverUser = await UserModel.findOne({
+      accounts: receiverAccount,
+    });
+    if (!receiverUser) {
+      await receiver.send(
+        "You were just praised in the TEC! It looks like you haven't activated your account... To activate use the `/praise-activate` command in the server."
```
kristoferlund, line 151: Configurable message text in settings.

src/bots/discord/commands/praise.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    await PraiseModel.create({
+      reason: reason,
+      giver: userAccount!._id,
+      sourceId: `DISCORD:${guild.id}:${interaction.channelId}`,
+      sourceName: `DISCORD:${encodeURI(guild.name)}:${encodeURI(
+        guildChannel?.name || ''
+      )}`,
+      receiver: receiverAccount!._id,
+      createdAt: interaction.createdAt,
```
kristoferlund, line 163: The `created_at` is auto created when document is created in mongo. Let's use that automatic date so we don't create confusion if server timezone is different from discord timezone (it most likely is).

src/bots/discord/deploy-commands.ts
```diff
@@ -1,38 +1,26 @@
-import { REST } from '@discordjs/rest';
-import { Routes } from 'discord-api-types/v9';
-import fs from 'fs';
+import { registerCommands } from './utils/registerCommands';
 
 const args = process.argv.slice(2);
-if (args.length != 2) {
-  console.log('deploy-commands accepts two arguments, token and client id ');
+if (args.length != 3) {
+  console.log('deploy-commands accepts three arguments, token, client id and guild id');
```
kristoferlund, line 5: We could skip reading arguments alltogether, right? And instead rely only on `.env`? When I created this script I didn't have access to env variables.

kristoferlund: Oh. Is this script needed at all even? You register commands on bot startup now, yes? If so, let's fry this file.

src/bots/discord/utils/praiseEmbeds.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const praiseErrorEmbed = (title: string, description: string) => {
+  return new MessageEmbed()
+    .setColor('#ff0000')
+    .setTitle(`ERROR: ${title}`)
+    .setDescription(description)
+    .setFooter({ text: 'PRAISE DID NOT REGISTER' });
+};
+
+export const praiseSuccessEmbed = (praised: string[]) => {
```
kristoferlund, line 11: I think we should make this message simpler and more like the one the current bot is using.

```
🥔 Vyvy-vi used /praise
Praise [BOT] today at 3:04 AM
@kristoferlund @someotherperson for this and that and a bunch of other stuff.
✅
```

kristoferlund (approved):
NICE WORK!

src/bots/discord/deploy-commands.ts
Vyvy-vi: yes, this can be removed :+1:

PR #82, opened 2022-02-18
System settings upgrade

nebs-dev:


mattyg:
I'll be offline most of today but will make sure to review tomorrow!

mattyg (commented):
Looks good, just two issues I wanted to note:

- Will there be a need to upgrade the schema of an existing databases? If so, we should include a migration script to update the settings schema but keeping their values. Perhaps using a migration tool like [migrate-mongoose](https://github.com/balmasi/migrate-mongoose)

- I noticed a UX quirk when saving settings. When I change an input value and click "Save Settings", the notice pops up saying that the setting was changed, but the input field displays the original unchanged value. After page refresh the input field displays the changed value as expected. This seems to occur consistently in text fields, but only inconsistently in number fields.

kristoferlund:
@mattyg 

> Will there be a need to upgrade the schema of an existing databases? If so, we should include a migration script to update the settings schema but keeping their values. Perhaps using a migration tool like [migrate-mongoose](https://github.com/balmasi/migrate-mongoose)

Yes! Made a new issue for migrations #109, @nebs-dev will be looking into that one.

> I noticed a UX quirk when saving settings. When I change an input value and click "Save Settings", the notice pops up saying that the setting was changed, but the input field displays the original unchanged value. After page refresh the input field displays the changed value as expected. This seems to occur consistently in text fields, but only inconsistently in number fields.

Fixed.

PR #114, opened 2022-02-28
Fix/improve speed praise score calculation

### LINKED ISSUE #96: Speed up calculation of praise score

mattyg:
Resolves #96 

These changes give ~10x performance improvement to the `findPeriodDetailsDto` function. There is likely room for further improvement if necessary. 

**Overview**
- run two db queries in `findPeriodDetailsDto` asynchronously unordered
- modify `calculateReceiverScores` to run calculation of quantifier scores asyncronously unordered, then sum afterwards
- modify `calculateReceiverScores` to run calculation of receiver scores asyncronously unordered
- add db indexes to praises table on `praises.receiver` and `praises.quantifications.quantifier` to speed up db queries used in findPeriodDetailsDto
-  ensure db settings are only queried once when running `calculateReceiverScores`
- add api dependency 'lodash' for sum helper function

**Deployment**
- indexes will be created automatically when server is launched, no need for a migration script.


kristoferlund (approved):
Ran it locally and as you say, 10x improvement at least. Running the queries in parallel of course! Great!!

PR #119, opened 2022-03-01
Seed users in different roles with random data

mattyg:
**Overview**
- seed specified number of users with `QUANTIFIER` role and fake data
- seed specified number of users with only `USER` role and fake data
- continue seeding a set of predefined users as before
- refactor seedData function for modularity


kristoferlund (approved):
Looks good!

PR #121, opened 2022-03-01
Various small fixes 2022-03-01

kristoferlund:
Fixed a few small bugs and closed the majority of frontend lint warnings. There is still some work left to do there as I used a fair amount of `// eslint-disable-next-line`. 

See commit messages for details.


packages/frontend/src/pages/Activate/ActivatePage.tsx
```diff
@@ -13,11 +13,12 @@ import { useRecoilValue } from 'recoil';
 import ActivateButton from './components/ActivateButton';
 import EthAccount from './components/EthAccount';
 
-const hasMetaMask = () => {
+const hasMetaMask = (): boolean => {
```
mattyg, line 16: Maybe `hasMetaMask` should be pulled out into utils since it's defined in 2 components.

packages/frontend/src/pages/Settings/SettingsPage.tsx
```diff
@@ -7,30 +7,30 @@ import {
...
+// ): ValidationErrors | Promise<ValidationErrors> => {
+//   const errors = {} as any;
 
-  // Currently no validation
+//   // Currently no validation
 
-  return errors as ValidationErrors;
-};
+//   return errors as ValidationErrors;
+// };
```
mattyg, line 24: Should this be deleted?

mattyg (commented):
Looks good. Appreciate you dealing with all those type definitions!

I definitely understand the need to comment out eslint warnings sometimes, but maybe we should make a ticket to revisit those down the road?

Also, why did you remove the tests?

packages/frontend/src/pages/Activate/ActivatePage.tsx
kristoferlund: ✅

kristoferlund:
> Looks good. Appreciate you dealing with all those type definitions!
> 
> I definitely understand the need to comment out eslint warnings sometimes, but maybe we should make a ticket to revisit those down the road?

Yes, we should definitely revisit the commented warnings. And also revisit which warnings we enable in the first place. Me and @nebs-dev discussed following the Google TypeScript Style Guide. We now enforce some of those rules / guidelines but by no means all.

https://google.github.io/styleguide/tsguide.html

> Also, why did you remove the tests?

The testing implementation is old, not updated and by no means finished. I'd like to have a renewed discussion quite soon about frontend testing - what do we test, how, etc. If we decide to keep going down the same path as before we'll bring the old code over to a new feature branch.

PR #128, opened 2022-03-02
Docs/metatags/favicon

kristoferlund:
Closes praise_frontend / # 126

- Adds share image 
- Adds favicons in various sizes
- Adds basic metatags and open graph tags


packages/frontend/public/index.html
```diff
@@ -7,9 +7,26 @@
...
     <meta
       name="description"
-      content="Web site created using create-react-app"
+      content="Acknowledge community contributions, build a culture of giving and gratitude."
     />
-    <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
+    <link
+      rel="apple-touch-icon"
+      sizes="180x180"
+      href="%PUBLIC_URL%/apple-touch-icon.png"
```
mattyg, line 15: Should we add `PUBLIC_URL=` to `.env.template` to make sure we don't forget to define the public url?

kristoferlund: I believe `PUBLIC_URL` should only be defined when app is not hosted at the web server root.

mattyg: i see, makes sense

PR #129, opened 2022-03-03
Fix/improve quantifier distribution

### LINKED ISSUE #95: Improve quantifier assign distribution

mattyg:
Resolves #95 

**Overview**
- adds 2 dependencies: `bin-packer` and `munkres-algorithm`
- modifies UX of AssignDialog to reflect new algorithm's definition of "meeting quantifier pool requirements"
- Reimplements the quantifier distribution algorithm as:
  1. [bin packing](https://en.wikipedia.org/wiki/Bin_packing_problem) the recipients into buckets, where each item's size is recipient.praiseCount, and the bucket size is `PRAISE_PER_QUANTIFIER`
  2. cloning the buckets `PRAISE_QUANTIFIERS_PER_PRAISE_RECEIVER` times, into one big list
  4. assignment of quantifiers to allowed buckets via [restricted hungarian assignment](https://en.wikipedia.org/wiki/Hungarian_algorithm)
  5. adding dummy quantifiers to represent bins that lack a compatible quantifier


kristoferlund:
Nice! Reading the wiki page made me feel a little less stupid not having come up with a good algorithm. Always feels good to realise others have struggled with the same problem.

mattyg:
Added a check at the end of `assignQuantifiersDryRun` to log that the expected number of quantification assignments are being accounted for. If not, the application throws a 500. I feel like its better for the app to fall over than to miss praise in a quantification assignment.

packages/api/src/period/controllers.ts
```diff
@@ -205,9 +188,28 @@ const assignQuantifiersDryRun = async (
...
+  const totalReceiverPraise = sum(receivers.map((r) => r.praiseIds.length));
+
+  // Run "First Fit" bin-packing algorithm on list of receivers
+  //    with a maximum 'bin' size of: PRAISE_PER_QUANTIFIER * tolerance
+  //    where each item takes up bin space based on its' praiseCount
+  const result: PackingOutput<Receiver> = firstFit(receivers, (r: Receiver) => r.praiseCount, maxPraisePerQuantifier);
+  const bins: Receiver[][] = [...result.bins, ...result.oversized.map((r) => [r])];
+
+  // Clone the bins for each redundant assignment (as defined by setting PRAISE_QUANTIFIERS_PER_PRAISE_RECEIVER)
+  const redundantAssignmentBins: Receiver[][] = flatten(range(quantifiersPerPraiseReceiver).map((i) => bins.slice()))
```
kristoferlund, line 209: Is this intentional?

image

kristoferlund (changes requested):
Getting an error message when I try to assign.

image

packages/api/src/period/controllers.ts
```diff
@@ -217,69 +219,98 @@ const assignQuantifiersDryRun = async (
...
+        availableQuantifiersSkipped.push(q);
+      }
     }
   }
 
-  // Trim any unassigned quantifiers from the returned pool
-  while (pool[pool.length - 1].receivers.length === 0) {
-    pool.pop();
+  // Convert object of quantifiers back to array & remove any unassigned
+  const poolAssignments: Quantifier[] = Object.values(quantifierPoolById)
```
kristoferlund, line 286: I'm getting error here. Adding type to Object.values resolves it. 

```
Object.values<Quantifier>
```

packages/api/src/period/controllers.ts
mattyg: oops - i think my eslint-vscode extension wasn't running. Fixed.

packages/api/src/period/controllers.ts
mattyg: Added. Was this a typescript error at launch? I'm not seeing it for some reason.

mattyg:
Okay, I refactored the section that assigns the quantifiers to the buckets of praise. It seems to be working reliably for me and feels like a much clearer and simpler implementation. I added another 3rd party dependency, to implement of the "(restricted Hungarian assignment)[https://en.wikipedia.org/wiki/Hungarian_algorithm"].

packages/api/src/period/controllers.ts
mattyg: nvm i think it was eslint

kristoferlund:
`verifyQuantifierPoolSize` no longer returns the required pool size but the additional needed members - `quantifierPoolDeficitSize`, right? 

If the pool size is too small I can easily find out the needed pool size by adding `quantifierPoolDeficitSize` to current pool size. But, if pool size is big enough, there is no way to tell how many will be assigned before pressing the button. Could that be added back in again? I think it is valuable to know for an admin who might want to tweak settings before a quant and see how pool size requirements change.

kristoferlund:
Looks good! 

I added `quantifierPoolSizeNeeded` to the `verifyQuantifierPoolSize`.

kristoferlund:
The distribution still becomes somewhat uneven though when a period has a quite large amount of praise (it seems).

image

I haven't tried to figure out what could be the cause.

Periods with smaller amount of praise seem to get more even distributions:

image

kristoferlund (approved):
Ok to merge this? I'd like it to be live before praise quant starts monday. The issue with distribution being slightly uneven with large  amounts of praise most likely won't affect the first quant period.

mattyg:
> Ok to merge this? I'd like it to be live before praise quant starts monday. The issue with distribution being slightly uneven with large amounts of praise most likely won't affect the first quant period.

Sorry I missed this. Take a look at my new PR, I think it should be stable now, despite some known issues I noted.

PR #141, opened 2022-03-07
Hotfix/quantifier assignment

mattyg:
**Overview**
- Reverts https://github.com/CommonsBuild/praise/pull/129/commits/506f4c0ee882f4b344f7765aa9f011c281cbaaf4
- fixes bug where quantifier assignment fails due to unknown error in `munkres-algorithm` dependency, removes dependency
- improves stability of quantifier assignment
- "smooths" quantifier assignment variation somewhat.

There are still two outstanding issues with the feature overall, but I think its fine for our immediate needs:
- If the quantifier pool size is near the minimum required quantifier pool size, you can't predict if there will be enough quantifiers, because you don't know the composition of the praise in relation to the quantifiers
- Because of how the bin-packing algorithm we're using works, there may be several buckets that are very "under-packed".  The number of buckets corresponds to setting `PRAISE_QUANTIFIERS_PER_PRAISE_RECEIVER`


kristoferlund (approved):
Nice!!

PR #146, opened 2022-03-09
Feat/renew expired jwt

### LINKED ISSUE #139: Renewal of expired JWT tokens

mattyg:
Resolves #139 
Implements jwt refresh tokens following this basic architecture: https://github.com/bezkoder/jwt-refresh-token-node-js
(A user with an expired accessToken, but an active refreshToken, can request a new accessToken with a later expiration date)

**Overview**
- add dependencies: `recoil-persist` (for automatic persistence of specified atom to localStorage), `recoil-nexus` (for calling recoil get / set outside of components), `axios-auto-refresh` (for handling refreshing tokens via axios instance).
- refactoring of Login component to improve flow & reliability
- refactoring of recoil axios request wrappers -- extracted into axios instances defined in `utils/api`. Recoil wrappers are kept, but simply call the axios instance, to limit the scope of this refactor.
- remove recoil auth query wrappers -- moved into `utils/auth`
- ensure expired tokens fail validation in api middleware
- logout user (redirect to `/login` and delete saved TokenSet) upon receiving a 401 response after refreshing tokens

### Changes to api's .env vars:
- `COOKIE_EXP` has been renamed `JWT_ACCESS_EXP` -- this defines the timespan of an access token
- `JWT_REFRESH_EXP` added -- this defines the timespan of a refresh token.

### Note about refactoring
- I kept all the api recoil selectors (ApiGet, ApiPost, etc.) in place to limit the scope of this PR 
- I think ideally, that layer of abstraction would be removed, and recoil query selectors would simply call the axios instances directly.
- I suspect this refactor may "go against the grain" of recoil design patterns. I'm not sure if I just don't fully understand "the recoil way", or if it is not well suited for this use case.


kristoferlund (changes requested):
Looks super solid! 

One bug has been introduced though. The app no longer fully reacts to switching of eth accounts. Like this: 
1. Log in 
2. Switch eth address in Metamask to address that have never logged in
3. You will now keep using the access token received in step 1

Expected functionality is: When switching eth address, use accesstoken for that address if there is one locally saved. Otherwise, redirect to login page.

packages/api/tsconfig.tsbuildinfo
```diff
@@ -1 +1 @@
...
...239,241,236,235,234,197,237,242,243,245,244,187,246,202,188,195,193,194,305,301,303,304,306,307,103,308,293,309,310,102,123,294,124,318,316,317,311,315,312,314,99,104,154,319,320,322,323,324,328,329,325,326,327,330,331,332,333,334,335,336,313,275,272,273,274,278,276,277,337,338,255,339,158,159,157,160,161,162,163,164,165,166,167,168,169,100,321,297,141,341,342,340,44,45,47,48,49,50,51,52,53,54,55,56,57,58,59,60,61,46,93,62,63,64,94,65,66,67,68,69,70,71,72,73,74,75,76,78,77,79,80,81,82,83,84,85,86,87,88,89,90,91,92,343,344,345,346,347,98,256,97,352,353,355,354,356,358,360,357,359,361,362,348,351,363,364,365,366,350,367,368,101,369,370,371,372,374,373,375,376,377,378,379,267,106,105,295,296,349,107,95,248,249,250,247,140,125,126,127,128,129,130,131,132,133,134,135,136,137,138,139,117,116,115,114,113,96,108,142,111,109,110,112,9,10,14,13,3,15,16,17,18,19,20,21,22,4,5,26,23,24,25,27,28,29,6,30,31,32,33,7,34,35,36,37,8,42,38,39,40,41,2,1,43,12,11,252,253,251,258,260,259,261,257,254,121,119,174,118,120,291,270,151,271,268,149,269,150,279,152,280,264,148,265,146,155,172,289,292,290,282,171,283,281,156,170,122,266,262,147,298,285,145,286,284,144,287,153,288,263,143,299]},"version":"4.6.2"}
```
kristoferlund, line 1: Should we really check in this file? Running `yarn workspace api build` produced no changes here locally until I deleted tsbuildinfo.

kristoferlund:
Also, pushed two tiny changes.

package.json
```diff
@@ -27,5 +27,8 @@
     "prettier": "^2.5.1",
     "typescript": "^4.5.5"
   },
-  "packageManager": "yarn@3.2.0"
+  "packageManager": "yarn@3.2.0",
+  "dependencies": {
+    "axios-auth-refresh": "^3.2.2"
```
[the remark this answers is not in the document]
mattyg: good catch thanks!

packages/api/tsconfig.tsbuildinfo
mattyg: Good catch -- I think it wasn't being ignored in my vscode git client until I added another line break. Fixed.

mattyg:
> Looks super solid!
> 
> One bug has been introduced though. The app no longer fully reacts to switching of eth accounts. Like this:
> 
> 1. Log in
> 2. Switch eth address in Metamask to address that have never logged in
> 3. You will now keep using the access token received in step 1
> 
> Expected functionality is: When switching eth address, use accesstoken for that address if there is one locally saved. Otherwise, redirect to login page.

Okay -- I think I resolved this. Now when switching your wallet address, all saved access / refresh tokens are deleted, and thus you are considered unauthenticated and directed to the login page. I don't think there should be a case where a user has two sets of access tokens stored simultaneously, so the user should always be unauthenticated and redirected to login upon switching wallet accounts.

I also resolved a few other minor issues:
- only request new nonce once for each signature
- no longer query /users/all after admin logout
- modify api auth middleware to respond with 401 if user does not exist, 403 is user forbidden from accessing content

kristoferlund:
> Okay -- I think I resolved this. Now when switching your wallet address, all saved access / refresh tokens are deleted, and thus you are considered unauthenticated and directed to the login page. I don't think there should be a case where a user has two sets of access tokens stored simultaneously, so the user should always be unauthenticated and redirected to login upon switching wallet accounts.

Yes, that is right. No need to be able to switch between more than one logged in account. It has been good during development but is definitely not a requirement for regular users since one user should be = one account.

kristoferlund (approved):

PR #150, opened 2022-03-13
Disallow active quantifier removal

### LINKED ISSUE #120: Disallow removal of active quantifiers from pool
If a quantifier is given some assignments for a period, an admin user can still remove their role as `QUANTIFIER` via the Quantifier Pool page. 

This would leave the data in an improper state, where the quantifier still has a bunch of assignments, but is unable to complete them.

We should implement a back-end check and front-end dialog to prevent removal of active quantifiers from the pool, or handle re-assigning their assignments.

mattyg:
Quick fix for #120

**Overview**
- throw error when removing quantifier if they are assigned to any praise in period with `period.status = 'QUANTIFY'`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


kristoferlund (approved):
Looks good!

PR #151, opened 2022-03-13
Ignore pound in praise dupe search

### LINKED ISSUE #144: Ignore "#" in search string when searching for duplicate praise
We should ignore the preceding '#' in the search string so it does not affect the results.

This was a cause of user confusion, see the screenshot / post below:

https://discord.com/channels/810180621930070088/932649711259766865/950486341500682300
image

mattyg:
Resolves #144


kristoferlund (approved):
👍👍

I added `trim()` to increase tolerance even more.

PR #154, opened 2022-03-14
Specify back link routes

mattyg:
Resolves https://github.com/commonsbuild/praise_frontend/issues/42

I implemented a different solution than specified in the above ticket. Rather than hide the BackLink component if the user was directed from an outside site based on browser history, the BackLink navigation path is specified explicitly. Lmk if the originally proposed solution is still preferred.

**Overview**
- Adds a "to" prop to the BackLink component
- Updates all uses of BackLink to specify 'to' prop

### DIFF
--- a/packages/frontend/src/navigation/BackLink.tsx
+++ b/packages/frontend/src/navigation/BackLink.tsx
@@ -1,12 +1,16 @@
 import { useHistory } from 'react-router-dom';
 
-const BackLink = (): JSX.Element => {
+interface BackLinkProps {
+  to?: string;
+}
+const BackLink = ({ to }: BackLinkProps): JSX.Element => {
   const history = useHistory();
 
   return (
     <div
       onClick={(): void => {
-        history.goBack();
+        if (to) history.push(to);
+        else history.goBack();
       }}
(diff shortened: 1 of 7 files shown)


kristoferlund (approved):
Using the `to` makes it less of a backlink. Take this example:

I as a quantifier click the link on the first page: "You can perform quantifications for the following periods". On the quantify page I press the backlink and end up not going back but instead navigating to quantification periods page. Unexpected behaviour. 

That said, I am happy to leave it as is for now. I'll try to involve a UI person quite soon to give this and a lot more an overhaul.

mattyg:
> "You can perform quantifications for the following periods". On the quantify page I press the backlink and end up not going back but instead navigating to quantification periods page. Unexpected behaviour.

My thought was the back button is really trying to indicate "returning to the previous level of the navigation hierarchy" like breadcrumbs. IMO the user already the browser's back button for returning to the previous page.

kristoferlund:
The unexpected behaviour happens when you navigate not according to the hierarchy, like when navigating "sideways" from period detail to quantification summary.

image

But as I said, I think it is totally ok to leave it as is for now and involve a UI person a little further on.

PR #152, opened 2022-03-13
github actions to build & lint PRs

mattyg:
Resolves #107 

- Note that we seem to be experiencing this issue in the github action `setup-node` https://github.com/actions/setup-node/issues/441

PR #163, opened 2022-03-15
Fix: sign button not working

kristoferlund:
The sign message button / activate button was unclickable. This does not seem to work:

```
onClick={void signLoginMessage}
```

I added a small click handler that in turn calls async function.


mattyg:
Oof my bad. I reviewed the #152 changes again and found they broke the SettingsPage and PeriodCreatePage forms too. 

For now I reverted the changes on those forms and instead disabled the eslint rule `@typescript-eslint/no-misused-promises`. I think there are implementation issues with both forms, as different bugs appear when wrapping `handleSubmit` in a sync function.

mattyg (approved):

Issue #165, opened 2022-03-16
JWT refresh not working


kristoferlund:

kristoferlund:
@mattyg 

There is some issue with JWT token refreshing. See below.

https://images.zenhubusercontent.com/438906728/39957e5e-17d4-4228-97f8-73ff1cf7f206/ska_rminspelning_2022_03_16_kl__09_03_23.mov

![Skärmavbild 2022-03-16 kl. 09.04.10.png](https://images.zenhubusercontent.com/5fdba670baa205320c5a291e/7656c87f-5905-4a48-859d-7f86ad3ab7b8)

mattyg:
I'm a bit confused by this ticket. I see the UX weirdness in the video -- which I'll take a look at. 

But are you saying the user was able to make authenticated requests after their refresh token expired? From the log it looks like the user did have their authentication refreshed, but the endpoint was called twice and the first attempt succeeded so the second failed. That seems fine (other than throwing a 404, I'll make it a 401 instead).

I haven't been able to reproduce this issue where the user is able to acquire a new access token after their refresh token has expired.

What are your api env variables `JWT_REFRESH_EXP` and `JWT_ACCESS_EXP` set to? Are you certain JWT_REFRESH_EXP of seconds had passed since login?

mattyg:
Note that Kristopher was missing the new env settings noted above. Closing.

PR #170, opened 2022-03-16
Improve refresh token error handling

mattyg:
Related to #165 

**Overview**
- modify `/auth/refresh` endpoint to throw a 401 if no user is found with matching refreshToken, rather than 404
- Reduce occurrences of 'Server did not respond' error toast by only displaying it if the request was successful, but there was no response.


kristoferlund (approved):
👍👍

PR #182, opened 2022-03-21
export average score fix for duplicate praise

nebs-dev:


packages/api/src/useraccount/types.ts
```diff
@@ -19,7 +19,7 @@ export interface UserAccountDocument extends UserAccount, mongoose.Document {}
 
 export interface UserAccountDto {
   _id: string;
-  user?: string;
+  user?: UserDocument;
```
kristoferlund, line 22: Mongoose documents shoould not be a part of a Dto object meant to be transferred over http. `user` should be `string`.

kristoferlund (changes requested):

kristoferlund:
Nice work!

PR #183, opened 2022-03-21
Allow scroll on duplicate dialog

### LINKED ISSUE #159: Allow background scroll while marking duplicate praise
While scrolling through the list of praise to quantify it can be quite lengthy - as a quantifier when I want to mark a praise duplicate of another instance of praise that might not be visible on screen, I have to exit the modal, scroll up or down, find the ID, scroll back down and find the praise I wanted to mark again.. 

We could make this a bit easier by allowing us to scroll the page while the mark duplicate praise modal is open. Does that make sense?

image

mattyg:
Resolves #159


kristoferlund (approved):
Nice!

PR #185, opened 2022-03-22
Settings labels & descriptions  + DB Migrations Setup

### LINKED ISSUE #109: DB migrations
From @mattyg  
> Will there be a need to upgrade the schema of an existing databases? If so, we should include a migration script to update the settings schema but keeping their values. Perhaps using a migration tool like migrate-mongoose

### LINKED ISSUE #166: Settings - Add label and description

mattyg:
Resolves #166 
Resolves #109

**Overview**
- add fields label & description to Settings model
- display label & description in frontend settings form
- add dependency `umzug` to api for managing db migrations
- setup db migrations management system
- create db migration to update initial settings with a default label & description
- make umzug cli available as command `yarn workspace api migrator`

Note that db migrations should work with a fresh db, or with the expected database schema before this PR (i.e. they can be safely run on a production database).


kristoferlund (changes requested):
Looks good! 

Migration of old settings is not handled though. We need that to upgrade the current TEC installation.

mattyg:
> Migration of old settings is not handled though. We need that to upgrade the current TEC installation.

Good call.  I just went ahead and implemented #109 here:

**Overview**
- added dependency `umzug`
- setup umzug for db migrations management
- run all pending migrations on server launch
- create db migration 1 to insert default settings (or skip if setting already exists), excluding label & description
- create db migration 2 to update default settings with default label & description

kristoferlund (approved):
Looks great!!

Issue #176, opened 2022-03-17
Make relevant parameters visible to quantifiers


mattyg:
It might be valuable to display relevant parameters to active quantifiers, at least `PRAISE_QUANTIFY_DUPLICATE_PRAISE_PERCENTAGE`

From a quantifier at the recent quantification review:
> I wanted to make sure duplicates had some value so I raised the score of the original contribution a little bit. If the original contribution is less than 10 then the duplicate could be rounded to 0. 

see https://docs.google.com/document/d/1bvpgt8YbDJxdrC3i3cEK9HJYwPwrqNiaUV8nFDyDPgc/edit#

PR #209, opened 2022-03-28
per period settings

### LINKED ISSUE #116: Ensure historical praise scores are accurate
If the setting `PRAISE_QUANTIFY_DUPLICATE_PRAISE_PERCENTAGE` is modified, it will cause historical praise score calculations to reflect the updated setting.

Do we want historical praise score calculations to change if settings are updated? If necessary, we can implement one of the solutions outlined below to ensure historical praise calculations remain static.

**Options**
- store praise score calculation in db, based on the settings at the time of calculation, rather than re-calculating on every fetch
- store settings history in db, calculate praise based on settings at the time of praise given, rather than current settings

### LINKED ISSUE #138: Settings tied to periods
See also #116

mattyg:
Resolves #138 
Resolves #116 
Improves #176 

**Deployment Steps**
- **add new api env variable `SERVER_URL` ( the internet-facing url of the api )**
- production data will be automatically migrated to support per-period settings

**Overview**
- adds required api env variable `SERVER_URL` 
- add frontend dependency `lodash`
- Per Period Settings feature:
  - all users can view period settings
  - users with role ADMIN can update period settings only if period status is 'OPEN'
  - periods are initially created with default period settings
- UI improvements
  - improve notices on QuantifierTable and ReceiverTable
  - 'disabled' option on SettingsForm to gray out all content / remove submit
- refactoring
  - frontend: extract form fields out of SettingsForm and into their own components, 
  - backend: workaround helper function `getPraisePeriod` to determine the period "associated" with a given praise instance 
  - move logic to transform setting string stored in database to "realized" value down to the database ORM layer with a mongoose virtual `setting.valueRealized`. Remove redundant implementations at all other layers.
  
  
**Per-period settings architecture**
- clone settingSchema to new periodSettingSchema
- add boolean field `periodOverridable` to settingSchema
- add ref field `period` to periodSettingSchema
- on period create, copy all from SettingModel with `periodOverridable=true` into new PeriodSettingModels
- clone setting controllers into new period setting controllers, limited to a specified period
- duplicate recoil settings atoms+selectors+queries for period settings
- refactor SettingForm to support 'disabled' ux, any arbitrary list of Setting or PeriodSettingDto
- convert PeriodSettingDto type to Setting type for usage by SettingForm to avoid larger refactor of setting types


kristoferlund:
Super!

- I made a small fix to praise export. It bypassed settings system and picked up value straight from env.
- Also, removed `font-bold` at a couple of places. Let's not use bold in "running text".
- Finally, adjusted an error message.

Merging, great work!

PR #213, opened 2022-03-30
Feat/praise forward UI

kristoferlund:
- Makes sure forwarder is returned by the api.
- Adds forwarder info to UI.

Skärmavbild 2022-03-30 kl  12 34 58

Skärmavbild 2022-03-30 kl  12 35 37


packages/api/src/praise/transformers.ts
```diff
@@ -99,8 +100,9 @@ const praiseDocumentToDto = async (
     quantifications: await quantificationListTransformer(quantifications),
     giver: userAccountTransformer(giver),
     receiver: userAccountTransformer(receiver),
-    createdAt: createdAt.toISOString(),
-    updatedAt: updatedAt.toISOString(),
+    forwarder: forwarder ? userAccountTransformer(forwarder) : undefined,
+    createdAt: createdAt ? createdAt.toISOString() : '',
+    updatedAt: updatedAt ? updatedAt.toISOString() : '',
```
mattyg, line 105: all mongoose models will always have createdAt & updatedAt defined -- so these conditionals are not necessary

mattyg (changes requested):
There's a couple eslint warnings and one change that I'm not sure about (see inline comments). Otherwise looks good!

packages/api/src/praise/transformers.ts
kristoferlund: Oh, yeah, I was debugging an error message and forgot to remove that.

PR #222, opened 2022-04-01
Fix/assign quantifiers only once

### LINKED ISSUE #101: Handle avatar images that don't load

mattyg:
Short-term workaround to reduce risk of https://github.com/commonsbuild/praise_frontend/issues/101
Resolves #101  -- I created a new ticket #223 to consider more "ideal" solutions.

**Overview**
- (merged `feat/per_period_settings` to avoid a tricky merge later)
- created utilily function `verifyAnyPraiseAssigned()` to check if any praise within a given period has already been assigned to quantifiers
- modify period controller `assignQuantifiers` to throw an error if any praise has been assigned for the period.


kristoferlund (approved):
Nice! And yes, lets make sure we have a more waterproof way to handle concurrent api calls etc in the future.

PR #224, opened 2022-04-01
Feat/sort recievers by score

### LINKED ISSUE #186: Sort receiver list on praise score

mattyg:
Resolves #186


packages/frontend/src/pages/Periods/Details/components/ReceiverTable.tsx
```diff
@@ -39,10 +40,22 @@ const ReceiverTable = (): JSX.Element | null => {
       ],
       []
     );
+    const data = period?.receivers
+      ? sortBy(period.receivers, [
```
kristoferlund, line 44: `react-table` has built in sorting. See a few lines below. 

But, perhaps we should try to move away from `react-table` where it is not actually motivated?

kristoferlund (commented):
Nice! Merging. 

Note to self/us, we should consider when to use and not to use `react-table`. 

On an updated user admin page for instance it feels totally motivated. There we might want to filter the view dynamically and move between "pages" of users - we only want to list xx number of users at the same time, community might consist of thousands of users.

On this page and others, it feels less motivated.. unless we want to add sorting symbols (▼ / ▲) for the columns. Then `react-table` makes sense again.

PR #226, opened 2022-04-02
Fix/mongodb clean

### LINKED ISSUE #136: yarn mongodb:clean command broken (auth)
After the MongoDB auth upgrade, this command has stopped working since the db can't be altered without proper credentials.
For now, we could manually add our dev environment credentials in the script, however, it would be neat if those can be parsed out of the `.env` file
Screenshot 2022-03-05 at 11 08 15 AM

mattyg:
**Overview**
- Resolves #136 
- (merges and adds support for feat/per_period_settings)


kristoferlund (approved):
Nice!

PR #241, opened 2022-04-05
Refactor Period endDate validators

### LINKED ISSUE #237: Error when changing endDate of one and only period
If there is only one period created, and an admin tries to update the period.endDate an error is thrown from one of the mongoose validator rules.

**Steps to reproduce**
- Create a single period
- edit the period date 
(throws an error)

mattyg:
**Overview**
- Resolves #237 
- Refactor period endDate validators for clarity, removing redundant logic.


kristoferlund (approved):
👍👍👍

PR #246, opened 2022-04-05
remove unused env vars

mattyg:
**Overview**
- Clone default settings from `.env.template` directly into initial migratation `00_initial_settings.ts`
- Removes default settings from .env.template


kristoferlund (approved):
👍👍👍

PR #257, opened 2022-04-07
Fix: Removed unused dependencies

kristoferlund:
I had a look at our dependencies. Turns out a lot of them where unused. 

Also in this PR:
- Removed cypress folder from frontend since it is currently unused and not updated


mattyg (approved):
looks good - thanks for cleaning these up

PR #266, opened 2022-04-09
Fix/ validation failure during assign quantifiers

### LINKED ISSUE #259: Modified date handling prevents assigning of quantifiers
> This might have to do with the date handling fix. If we handle dates slightly different, that might mean praise from a previous period now belongs to this. And that in turn causes assign to fail. You are not allowed to assign quantifiers if there is praise in the period that already have quantifiers assigned.

image

mattyg:
Resolves #259 

The issue is caused by a mongoose validation error being improperly thrown due to changes from #241. (not related to timezone handling, nor the implicit relation of praise - date range). 

Running assignQuantifiers in v0.0.3 once should put the data in the following "dirty" state:
- Quantifiers *are* assigned properly for that period
- But the status of that period is 'OPEN', when it should be 'QUANTIFY'

However, reviewing the production database:
- There is only one period open with the status 'QUANTIFY'
- There are no praise created after that period's endDate with quantifications assigned (i.e. no results to the following query `db.praises.find({createdAt: {$gt: ISODate("2021-12-26T23:59:59.999Z")}, quantifications: {$not: {$size: 0}}})`)

I believe the production data is in a clean state and after this fix we should be able to assign quantifiers properly.

### DIFF
--- a/packages/api/src/period/controllers.ts
+++ b/packages/api/src/period/controllers.ts
@@ -143,7 +143,8 @@ export const update = async (
   }
 
   if (endDate) {
-    if (!isPeriodLatest(period))
+    const latest = await isPeriodLatest(period);
+    if (!latest)
       throw new BadRequestError('Date change only allowed on last period.');
 
     if (period.status !== PeriodStatusType.OPEN)
@@ -461,7 +462,10 @@ export const assignQuantifiers = async (
   );
 
   await PraiseModel.bulkWrite(bulkQueries);
-  await PeriodModel.updateOne({_id: period._id}, {$set: {status: PeriodStatusType.QUANTIFY}});
+  await PeriodModel.updateOne(
+    { _id: period._id },
+    { $set: { status: PeriodStatusType.QUANTIFY } }
+  );
 
   const periodDetailsDto = await findPeriodDetailsDto(periodId);
   res.status(StatusCodes.OK).json(periodDetailsDto);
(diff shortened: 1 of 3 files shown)


kristoferlund (approved):
I removed this check in `close()`:

```
if (period.status !== PeriodStatusType.QUANTIFY)
    throw Error('Only Periods with status QUANTIFY can be closed');
```

One might want to close a period without doing a quantification first.

kristoferlund:
> However, reviewing the production database:
> 
> * There is only one period open with the status 'QUANTIFY'
> * There are no praise created after that period's endDate with quantifications assigned (i.e. no results to the following query `db.praises.find({createdAt: {$gt: ISODate("2021-12-26T23:59:59.999Z")}, quantifications: {$not: {$size: 0}}})`)
> 
> I believe the production data is in a clean state and after this fix we should be able to assign quantifiers properly.

I cleaned the database manually on Friday to allow Mitch to start the quantification. All looked good except that period still had `OPEN` status.

mattyg:
> I removed this check in `close()`:
> 
> ```
> if (period.status !== PeriodStatusType.QUANTIFY)
>     throw Error('Only Periods with status QUANTIFY can be closed');
> ```
> 
> One might want to close a period without doing a quantification first.

On the front-end it's not possible to close a period from the 'OPEN' status (the 'close' button does not appear).

I think if we want users to be able to close a period from the 'OPEN' status we should modify the front-end to support it. Otherwise we should keep this check to reflect the current business logic.

PR #270, opened 2022-04-10
Fix/262/period selectmenu error

### LINKED ISSUE #262: "/admin announce" don't work
The command crashes trying to send PeriodSelectMenu, when the "All assigned quantifiers" or "All unfinished quantifiers" options are selected.
The command works as expected when selecting "All users" or "All quantifiers".

Vyvy-vi:
fixes #262 

The `/admin announce` command errors out for the "All assigned quantifiers" and "All unfinished quantifiers" options because the bot didn't account for a case when there are no periods with the `QUANTIFY` state.

This PR adds a fix for this condition. (Though, maybe we need an up[date to this text...)


kristoferlund (approved):
👍👍👍

Issue #99, opened 2022-02-23
Centralise Db connection params


kristoferlund:
The database connection parameters currently have to be duplicated in

```
/.env
/api/.env
/discord-bot/.env
```

Those params should only be in `/.env`.

Multiple calls to dotenv are allowed. Include top level `.env` from `/api` and `/discord-bot`.

Issue #267, opened 2022-04-09
Check for required env variables on startup


mattyg:
On startup: check that all required env variables are defined in .env -- otherwise throw an error and exit.

This should help prevent deployments without required env variables

For both:
- [ ] discord bot
- [ ] api

mattyg:
Resolved by #288

PR #278, opened 2022-04-11
feat/dockerisation

### LINKED ISSUE #261: Serving of uploaded images don't work

kristoferlund:
### MongoDb
- Created a separate package for MongoDB

### Docker
- Created "Dockerised" versions of api, discord-bot and frontend
- Frontend is served using Caddy!
  - Handling of SSL certificates is completely automated
  - Added reverse proxy for `/uploads`, eliminating the need for the frontend to call backend directly for the image uploads

### `ENV` handling:
- Fixes: #99 
- Added a few assumptions to have to handle less env variables
  - No more `FRONTEND_URL`, `PORT`, `HOST`, `SERVER_URL`. Not needed anymore. 
  - Instead using one single variable `SERVER` for the server domain name, either `localhost` or external domain.
  - Running frontend locally assumes using default port 3000
  - Api runs on port 8088, currently not configurable. That can of course be added if need arises.
- Centralised database env variables. These are now only stored in root level `.env`
- Unified naming of database env variables:
  - `MONGO_DB_USER` > `MONGO_USER` 
  - `MONGO_DB_PASSWORD` >  `MONGO_PASSWORD`
  - `REACT_APP_BACKEND_URL` > `REACT_APP_API_URL`

### Updated root `package.json`
  - Docker compose build scripts to build the individual Docker images
  - Start scripts for all services  
  - `yarn start` builds and starts all services

### Image settings
- Move to store only the uploaded filename in the database, not the path included.
- Fixes #261 

### Misc
- Excluded `api/src/scripts` from api build since it caused some kind of collision
- Updated install instructions
  - The "[minimal install instruction](packages/docs/install-praise-on-digital-ocean.md)" now does not include instructions to setup users and configure firewalls
  - Firewall and user instructions now placed in separate docs
 - Added frontend utility function `imageFullPath`


README.md
```diff
@@ -28,21 +30,64 @@ nvm use
...
+Copy and rename `.env`.
+
+- `DISCORD_TOKEN` - Your bot's discord token generated via the Discord Developer Portal. You'll need to invite the same bot to your server with the link - `https://discord.com/api/oauth2/authorize?client_id=<client-id>&permissions=378561611840&scope=bot%20applications.commands` (replace `<client-id>` with your bot's client ID), and with the SERVER MEMBERS and MESSAGE CONTENT Intents enabled.
+- `DISCORD_CLIENT_ID` - Your bot's discord client ID, which can be found in the Application settings on Discord Developer Portal
+- `DISCORD_GUILD_ID` - The ID of the server in which you are using the bot. (this can be found by enabling developer mode in Discord, right clicking on te server icon and clicking "Copy Id").
+
+```
+DISCORD_TOKEN=
+DISCORD_CLIENT_ID=
+DISCORD_GUILD_ID=
```
mattyg, line 78: I think the discord bot setup may be incomplete -- it keeps crashing for me with the following error:

```
discord-bot-praise | /usr/src/node_modules/discord.js/src/client/websocket/WebSocketManager.js:250
discord-bot-praise |         throw new Error(WSCodes[error.code]);
discord-bot-praise |               ^
discord-bot-praise | Error [DISALLOWED_INTENTS]: Privileged intent provided is not enabled or whitelisted.
discord-bot-praise |     at WebSocketManager.createShards (/usr/src/node_modules/discord.js/src/client/websocket/WebSocketManager.js:250:15)
discord-bot-praise |     at async Client.login (/usr/src/node_modules/discord.js/src/client/Client.js:254:7) {
discord-bot-praise |   [Symbol(code)]: 'DISALLOWED_INTENTS'
discord-bot-praise | }
```

I think the discord bot may also need scopes for `guilds` and `guilds.members.read`?

README.md
```diff
@@ -28,21 +30,64 @@ nvm use
...
+```
+
+#### `/packages/frontend/.env.template`
+
+Copy and rename `.env.development`.
```
mattyg, line 83: Any reason not to just copy the env file to `.env`? Rather than `.env.development`?

packages/api/src/index.ts
```diff
@@ -1,17 +1,10 @@
-import * as dotenv from 'dotenv';
+import './loadEnv';
```
mattyg, line 1: :+1: :+1:

packages/api/src/index.ts
```diff
@@ -1,17 +1,10 @@
...
 
-const load = dotenv.config({ path: path.join(__dirname, '..', '/.env') });
-if (load.error) {
-  logger.err(load.error.message);
-  throw load.error;
-}
-
 // Start the server
-const port = Number(process.env.PORT || 3000);
+const port = 8088;
```
mattyg, line 7: Same note as above -- I really don't like overriding env settings with static values, conditionally.

packages/api/src/settings/entities.ts
```diff
@@ -39,8 +39,6 @@ export function getGenericSettingValueRealized(
     normalizedValue = this.value
       .split(',')
       .map((v: string) => Number.parseInt(v.trim()));
-  } else if (this.type === 'Image') {
-    normalizedValue = `${process.env.SERVER_URL as string}/${this.value}`;
   } else {
```
mattyg, line 42: I think it is preferable to define the "realized" image url via the server, and just load the specified url in the client.

It should be under the server's control where images are hosted, and under what path.  The client shouldn't need to know anything about our image hosting -- just the url to fetch.

That allows us to make sure every client is accessing images from the proper place in case anything needs to be moved down the road.

packages/api/src/shared/functions.ts
```diff
@@ -90,14 +90,15 @@ export const upload = async (req: Request, key: string): Promise<string> => {
...
+    return filename;
   } catch (e) {
     console.log('ERROR:', e);
     throw new InternalServerError('File upload failed.');
   }
 };
 
-export const removeFile = async (filepath: string): Promise<void> => {
+export const removeFile = async (filename: string): Promise<void> => {
+  const filepath = `uploads/${filename}`;
```
mattyg, line 101: Do we want to pull out the definition of the upload directory `uploads/` into env configuration?  (or else just defined as a constant in this file)

packages/discord-bot/src/handlers/activate.ts
```diff
@@ -24,21 +24,24 @@ export const activationHandler = async (
...
+  if (!server || server === undefined) {
     await interaction.reply(
-      'ERROR: `FRONTEND_URL` not defined in environment variables. Contact praise admin'
+      'ERROR: `SERVER` not defined in environment variables. Contact praise admin'
     );
     return;
   }
+  const frontendUrl = `http://${
+    process.env.NODE_ENV === 'production' ? server : `${server}:3000`
+  }`; // Assume default craco port 3000 if not in production
```
mattyg, line 36: Same note as above. I personally think that server url & port should remain configurable for each service via the `env` files. 

We can also include a partially completed "default" `env` file for development vs production environments.

packages/frontend/src/components/form/ImageFileInput.tsx
```diff
@@ -15,9 +16,9 @@ const ImageFileInput = (name: string, src: string | undefined): JSX.Element => {
 
           <div className="mt-2">
             {src ? (
-              <img src={src} className="w-auto h-48 block" />
+              <img src={imageFullPath(src)} className="block w-auto h-48" />
```
mattyg, line 19: Same note as above with `getGenericSettingValueRealized` -- I think the previous architecture is preferable (where the image's full url is defined by the server).

packages/frontend/src/navigation/Nav.tsx
```diff
@@ -34,7 +35,7 @@ export default function Nav(): JSX.Element {
               <Link to="/">
                 {logoSetting && (
                   <img
-                    src={logoSetting.valueRealized as string}
+                    src={imageFullPath(logoSetting.valueRealized as string)}
```
mattyg, line 38: Same note as above with getGenericSettingValueRealized -- I think the previous architecture is preferable (where the image's full url is defined by the server).

packages/frontend/src/utils/file.ts
```diff
@@ -7,3 +7,8 @@ export const saveLocalFile = (data: Blob, fileName: string): void => {
   a.click();
   window.URL.revokeObjectURL(url);
 };
+
+export const imageFullPath = (filename: string): string => {
```
mattyg, line 11: Same note as above with getGenericSettingValueRealized -- I think the previous architecture is preferable (where the image's full url is defined by the server).

mattyg (changes requested):
Take a look at my inline notes and let me know what you think. 

I also noticed we need to rebuild the docker images when updating an `env` variable. It might be nice if we had a "local" version of docker-compose.yml to read the local .env, rather than from the docker volume.

README.md
kristoferlund: I think current setup of React requires `.env.development` or `.env.production`. I would have preferred removing frontend env completely but React needs separate variables beginning with `REACT_APP`.

packages/api/src/index.ts
kristoferlund: One of my concerns was making switching between the three modes of running as easy as possible:
1. Docker - production
2. Docker - development
3. No docker - development

I'll try to add PORT support back in.

README.md
mattyg: It's kinda confusing but it looks like with react-scripts `.env` is always the "default" config, with others overriding it conditionally:
https://create-react-app.dev/docs/adding-custom-environment-variables/#what-other-env-files-can-be-used

packages/api/src/database/connection.ts
```diff
@@ -1,9 +1,11 @@
 import mongoose, { ConnectOptions } from 'mongoose';
 
-const connectDatabase = async (): Promise<typeof mongoose> => {
+const connectDatabase = async (
+  hostOverride?: string
+): Promise<typeof mongoose> => {
   const username = process.env.MONGO_USERNAME || '';
   const password = process.env.MONGO_PASSWORD || '';
-  const host = process.env.MONGO_HOST || '';
+  const host = hostOverride || process.env.MONGO_HOST || '';
```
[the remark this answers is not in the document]
mattyg: Thinking through it more -- I realized there is probably a good case for overriding env values for different environments: 
- it reduces the risk that devs accidentally run a dangerous command in the wrong environment
- it makes changing the setup for different environments simpler

Ideally we should only override env variables based only on the `NODE_ENV` - and all the logic describing that would be clear and kept in one place (a single conditional statement). Then we can override the `NODE_ENV` in individual package.json script commands.

For now though, I think this is fine. Let's just open another ticket for the above as a refactor. @kristoferlund

packages/api/src/index.ts
mattyg: I was thinking another option would be to replace the mongo config variables with a single MONGO_URI.

mattyg: I was thinking another option would be to replace the mongo config variables with a single MONGO_URI.

README.md
kristoferlund: Yes, the discord bot needs additional setup. There is more info in one of the doc pages. The Discord setup should get its own page, I'll make an issue about that.

README.md
kristoferlund: Ah, nice, didn't know that! Will change then.

packages/api/src/database/connection.ts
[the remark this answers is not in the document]
kristoferlund: I have seen this being referenced here and there: https://12factor.net/

Haven't read it myself yet though. One thing they seem to advocate for is maintaining separate complete env files for different environments rather than overriding etc.

Perhaps I'll just remove the override for now because like we discussed - the scripts can be run from inside the Docker container and use the same env as the production environment.

packages/api/src/index.ts
kristoferlund: Above is not for mongo but the api port. But yeah, perhaps we'll switch to a single MONGO_URI as well.

packages/frontend/src/utils/file.ts
kristoferlund: Fixed

packages/frontend/src/navigation/Nav.tsx
kristoferlund: Fixed

packages/frontend/src/components/form/ImageFileInput.tsx
kristoferlund: Fixed

packages/discord-bot/src/handlers/activate.ts
kristoferlund: Fixed

packages/api/src/shared/functions.ts
kristoferlund: Fixed

kristoferlund:
## `ENV` handling
- Moved all env variables to one global file
- Packages no longer read the env file themselves but are fed the env using `env-cmd`
- No env variables are copied while building the Docker image. Changing an env var and restarting the container now works.'
- Added back support to set frontend and api port. Frontend port can only be set during development, in production it is locked to `80 / 443`.

## Image handling
- Removed the need for the frontend to make assumptions as to where images are hosted. Now, path delivered by api is law.

## Misc
- Removed ability to override host settings while running scripts. Scripts should run ok when run inside the Docker container-

.env.template
```diff
@@ -1,4 +1,47 @@
...
+DISCORD_TOKEN=
+DISCORD_CLIENT_ID=
+DISCORD_GUILD_ID=
+
+## LOGGING ##
+JET_LOGGER_MODE=CONSOLE
+JET_LOGGER_FILEPATH=jet-logger.log
+JET_LOGGER_TIMESTAMP=TRUE
+JET_LOGGER_FORMAT=LINE
+
```
mattyg, line 47: I don't think all the env variables for 3 applications should be kept in a single file. What if we want some variables to only be available to a single application and not others (like secret keys)?

Only the shared env variables should be in the shared .env files. Application-specific env variables should be within their own workspace. I think that was the correct approach.

mattyg: You can specify multiple env files for each docker application. See https://github.com/docker/compose/issues/6392#issuecomment-444265127

packages/api/package.json
```diff
@@ -8,14 +8,14 @@
...
     "lint": "eslint . --ext .ts",
-    "start": "ts-node -r tsconfig-paths/register --transpile-only ./src",
-    "start:dev": "nodemon ./src",
-    "test": "ts-mocha --paths -p tsconfig.json src/**/*.spec.ts",
-    "import-praise": "ts-node -r tsconfig-paths/register src/scripts/import-praise.ts",
-    "not-yet-activated": "ts-node -r tsconfig-paths/register src/scripts/not-yet-activated.ts",
-    "replace-quantifier": "ts-node -r tsconfig-paths/register src/scripts/replace-quantifier.ts",
-    "migrator": "ts-node -r tsconfig-paths/register src/scripts/umzug-cli.ts",
-    "drop-database": "ts-node -r tsconfig-paths/register src/scripts/drop-database.ts"
+    "start": "env-cmd --silent -f ../../.env ts-node -r tsconfig-paths/register --transpile-only ./src",
```
mattyg, line 11: Why did you take this approach to use `env-cmd` rather than reading env files in each workspace with `dotenv`?

mattyg:
> Added a few assumptions to have to handle less env variables
> * No more `FRONTEND_URL`, `PORT`, `HOST`, `SERVER_URL`. Not needed anymore.
> * Instead using one single variable `SERVER` for the server domain name, either `localhost` or external domain.
> * Running frontend locally assumes using default port 3000
> * Api runs on port 8088, currently not configurable. That can of course be added if need arises.

I think we should revert these changes. All seperate application settings should be configurable independently. We should be able to deploy the frontend and backend anywhere, independently, without modifying the codebase.

packages/api/src/scripts/import-praise.ts
```diff
@@ -100,7 +100,7 @@ const importPraise = async (
   process.exit();
 };
 
-connectDatabase().then(() => {
+connectDatabase('localhost').then(() => {
```
mattyg, line 103: Please see my changes to this file in #289 -- I think it is a cleaner approach to per-environment overriding of env variables.

packages/api/src/settings/entities.ts
```diff
@@ -40,7 +40,8 @@ export function getGenericSettingValueRealized(
       .split(',')
       .map((v: string) => Number.parseInt(v.trim()));
   } else if (this.type === 'Image') {
-    normalizedValue = `${process.env.SERVER_URL as string}/${this.value}`;
+    const apiURL = process.env.REACT_APP_API_URL;
+    normalizedValue = apiURL ? `${apiURL}/${this.value}` : `/${this.value}`;
```
mattyg, line 44: This conditional is unnecessary. We should just require the env variable to be defined via `env-required.ts`

packages/api/tsconfig.json
```diff
@@ -28,5 +28,6 @@
     }
   },
   "include": ["src/**/*.ts", "test/**/*.ts", "src/**/*.json"],
+  "exclude": ["src/scripts/*.ts"],
```
mattyg, line 31: This exclude is not necessary - the build errors I'm getting from those files look accurate and should be fixed.

packages/discord-bot/src/handlers/activate.ts
```diff
@@ -24,21 +24,32 @@ export const activationHandler = async (
...
+  const port = process.env.FRONTEND_DEVELOPMENT_PORT;
+
+  if (!server || server === undefined) {
+    await interaction.reply(
+      'ERROR: `SERVER` not defined in environment variables. Contact praise admin'
+    );
+    return;
+  }
+  if (!port || port === undefined) {
     await interaction.reply(
```
mattyg, line 37: The checks for env variables are no longer necessary due to #267

packages/discord-bot/src/handlers/activate.ts
```diff
@@ -24,21 +24,32 @@ export const activationHandler = async (
...
-      'ERROR: `FRONTEND_URL` not defined in environment variables. Contact praise admin'
+      'ERROR: `FRONTEND_DEVELOPMENT_PORT` not defined in environment variables. Contact praise admin'
     );
     return;
   }
+  const frontendUrl = `http://${
+    process.env.NODE_ENV === 'production' ? server : `${server}:${port}`
+  }`;
```
mattyg, line 44: This conditional should not be necessary. We can change the env variable FRONTEND_URL to include a port if desired.

packages/frontend/Caddyfile
```diff
@@ -0,0 +1,14 @@
+{$SERVER} {
+  handle /api* {
+		reverse_proxy api:{$API_PORT}
+	}
+  handle /uploads* {
+		reverse_proxy api:{$API_PORT}
```
mattyg, line 6: Whats the point of splitting up /api and /uploads if they're both just reverse proxies to the express backend? Why not just send all traffic to the express backend and let it do the routing?

packages/frontend/package.json
```diff
@@ -63,7 +64,7 @@
   },
   "scripts": {
     "build": "craco build",
-    "start": "TAILWIND_MODE=watch craco start",
+    "start": "PORT=$(grep FRONTEND_DEVELOPMENT_PORT ../../.env | cut -d '=' -f2) && TAILWIND_MODE=watch env-cmd -f ../../.env craco start",
```
mattyg, line 67: We should not override env variables in command except for NODE_ENV.  Also this is a crazy way to manually parse an env file that should not be necessary.

packages/frontend/src/utils/api.ts
```diff
@@ -45,17 +45,18 @@ const handleErrors = (err: AxiosError): void => {
   }
 };
 
+const apiBaseURL = process.env.REACT_APP_API_URL
+  ? `${process.env.REACT_APP_API_URL}/api`
+  : '/api';
```
mattyg, line 50: We shouldn't need this conditional -- simply require the env variable be defined via `env-required.ts`

mattyg (changes requested):
See the inline comments above.

packages/api/package.json
kristoferlund: When using Docker, we don't want to build on the assumption there being env files. We don't want to place copies of any env files inside the built Docker images. Instead we want the env to be supplied to the container at runtime. In `docker-compose.yml` this is achieved with the `env-file` directive. 

When running the code outside of Docker, env has to be supplied in another fashion, in this case I chose `env-cmd`.

kristoferlund:
> > Added a few assumptions to have to handle less env variables
> > 
> > * No more `FRONTEND_URL`, `PORT`, `HOST`, `SERVER_URL`. Not needed anymore.
> > * Instead using one single variable `SERVER` for the server domain name, either `localhost` or external domain.
> > * Running frontend locally assumes using default port 3000
> > * Api runs on port 8088, currently not configurable. That can of course be added if need arises.
> 
> I think we should revert these changes. All seperate application settings should be configurable independently. We should be able to deploy the frontend and backend anywhere, independently, without modifying the codebase -- just by changing env variables.

This is now the case again. What do you miss being able to configure? These ENV settings allow for full customisation:

```
SERVER
API_PORT
REACT_APP_API_URL
FRONTEND_DEVELOPMENT_PORT
```

packages/api/src/scripts/import-praise.ts
kristoferlund: 👍 ok great, any action needed on my part?

packages/api/src/settings/entities.ts
kristoferlund: Leaving this setting empty is allowed when running in production - when running the api on the same domain as the frontend. We could leave out the conditional though and always require the full url to the api to be in that variable even in the case of running on the same domain. What do you think?

packages/api/tsconfig.json
kristoferlund: 👍

mattyg:
> > > Added a few assumptions to have to handle less env variables
> > > 
> > > * No more `FRONTEND_URL`, `PORT`, `HOST`, `SERVER_URL`. Not needed anymore.
> > > * Instead using one single variable `SERVER` for the server domain name, either `localhost` or external domain.
> > > * Running frontend locally assumes using default port 3000
> > > * Api runs on port 8088, currently not configurable. That can of course be added if need arises.
> > 
> > 
> > I think we should revert these changes. All seperate application settings should be configurable independently. We should be able to deploy the frontend and backend anywhere, independently, without modifying the codebase -- just by changing env variables.
> 
> This is now the case again. What do you miss being able to configure? These ENV settings allow for full customisation:
> 
> ```
> SERVER
> API_PORT
> REACT_APP_API_URL
> FRONTEND_DEVELOPMENT_PORT
> ```

:face_with_spiral_eyes: Sorry my mistake. Should we rename SERVER to HOST?

packages/discord-bot/src/handlers/activate.ts
kristoferlund: Gone.

packages/discord-bot/src/handlers/activate.ts
kristoferlund: Currently we don't need `FRONTEND_URL` at all. The following variables allow for full customisation: 

```
SERVER
API_PORT
REACT_APP_API_URL
FRONTEND_DEVELOPMENT_PORT
```

The likelihood of us wanting to serve the production frontend on anything else than standard port 80/443 is veery small. Or am I missing something?

packages/frontend/Caddyfile
kristoferlund: What you mean?

Everything that is not `/api` and `/uploads` is the frontend and being directly served by Caddy.

One option though would be to let Express serve the frontend as well and have one less Docker image in total.

packages/frontend/package.json
kristoferlund: Ok. Slightly hacky, agreed. :) This came from my want to join all ENV in one file. 

We could move some stuff back to package individual env files. But, then we need to find a solution to this problem (`env-cmd` I believe only supports reading env from one file):

1. Read top level .env
2. Read package level .env
3. Merge and feed env to package on start

packages/frontend/src/utils/api.ts
kristoferlund: See comment above.

.env.template
kristoferlund: Yes, not as straight forward though specifying more than one env file outside of Docker. One alternative is to let packages load env as before, except when running inside Docker. 

How to know when running inside Docker? There is a library for that: https://www.npmjs.com/package/is-docker

kristoferlund: > The env variables for 3 applications should not be kept in a single file. What if we want some variables to only be available to a single application and not others (like secret keys)?
> 
> Only the shared env variables should be in the shared .env files. Application-specific env variables should be within their own workspace. I think that was the "best practice" approach.

Yeah, I intentionally disregarded the logger settings. Jet logger producing log files inside the Docker container is not ideal anyway, a revisit to the logging setup is needed. Let's make a separate issue for that one.

packages/api/src/settings/entities.ts
mattyg: Yeah I think always requiring the full url is preferable.

packages/api/package.json
mattyg: > When using Docker, we don't want to build on the assumption there being env files. We don't want to place copies of any env files inside the built Docker images. Instead we want the env to be supplied to the container at runtime. In `docker-compose.yml` this is achieved with the `env-file` directive.
> 
> When running the code outside of Docker, env has to be supplied in another fashion, in this case I chose `env-cmd`.

Hmmm - okay yeah that makes sense. Another option could be to revert back to reading env files with `dotenv` and use a docker volume to make the external `.env` file available within the container. But I can't think of any reason why this wouldn't be fine too.

packages/discord-bot/src/handlers/activate.ts
mattyg: > Currently we don't need `FRONTEND_URL` at all. The following variables allow for full customisation:
> 
> ```
> SERVER
> API_PORT
> REACT_APP_API_URL
> FRONTEND_DEVELOPMENT_PORT
> ```
> 
> The likelihood of us wanting to serve the production frontend on anything else than standard port 80/443 is veery small. Or am I missing something?

It looks like `FRONTEND_DEVELOMENT_PORT` is only used in this one place -- to define the `frontendUrl` value. So why not just specify the frontend url directly with an env variable?

Then the protocol, host, and domain can all be defined within that variable as needed -- allowing for maximum flexibility to deploy the frontend anywhere without needing to modify the codebase.

packages/frontend/Caddyfile
mattyg: Ah right okay. Another options would be to use two seperate Caddy hosts: one that is *just* a reverse proxy to the express server, and another that *just* serves the static frontend. That would allow us to define the frontend and api domains independently.

But I think this approach is fine.

packages/frontend/package.json
mattyg: It looks like you can just call `env-cmd` twice to merge two files: 

```
    "start": "TAILWIND_MODE=watch env-cmd -f ../../.env env-cmd -f .env --silent craco start",
```

From this comment: https://github.com/toddbluhm/env-cmd/issues/148#issuecomment-652807598

packages/api/src/scripts/import-praise.ts
mattyg: Yeah `connectDatabase` no longer takes any parameters - we're getting build errors due to calling it with a single parameter: `connectDatabase('localhost')`.

But also #289 changes the connectDatabase implementation -- we may need to carefully merge.

.env.template
mattyg: > Yeah, I intentionally disregarded the logger settings. Jet logger producing log files inside the Docker container is not ideal anyway, a revisit to the logging setup is needed. Let's make a separate issue for that one.

Hmm -- you could setup a docker volume to mount a path in system filesystem with the docker container's filesystem. See https://docs.docker.com/storage/volumes/#use-a-volume-with-docker-compose

packages/api/src/settings/entities.ts
kristoferlund: Ok, done.

.env.template
kristoferlund: I believe the output from jet-logger made to stdout is caught by Docker and stored outside the container. Let's investigate that before looking into other solutions.

packages/frontend/.env.template.production
```diff
@@ -0,0 +1,8 @@
+
+## FRONTEND ##
+
+# Full URL to host where API is running
+REACT_APP_SERVER_URL=https://localhost
+
+# Port number used when running frontend for development, outside of Docker
+PORT=3000
```
mattyg, line 8: Do you think we need both the `.env.template.production` and `.env.template.development` files? Would just one template suffice?

.env.template.production
```diff
@@ -0,0 +1,29 @@
...
+## DATABASE ##
+# Running through Docker: MONGO_HOST=mongodb
+# Running outside Docker: MONGO_HOST=localhost
+MONGO_HOST=mongodb
+MONGO_DB=praise_db 
+MONGO_PORT=27017
+MONGO_INITDB_ROOT_USERNAME=praiseDbRootUsername
+MONGO_INITDB_ROOT_PASSWORD=[any password]
+MONGO_USERNAME=praiseDbUsername
+MONGO_PASSWORD=[any password]
```
mattyg, line 29: Do you think it's necessary to have both `.env.template.production` and `.env.template.development`? IMO the comments make clear how to configure the env file so both templates are not necessary.

.env.template.development
```diff
@@ -0,0 +1,29 @@
...
+# Use HOST=localhost for development
+HOST=localhost
+
+# When running in development, API is accessed on this port, default: localhost:8088
+# In production, this port is used internally only since api is accessed
+# on $HOST/api 
+API_PORT=8088
+
+# Full URL to the host (and optionally port) where the API is running
+SERVER_URL=http://localhost:8088
```
mattyg, line 15: It looks like `SERVER_URL` is only used in the api. Should it be moved to `api/.env`?

.env.template.development
```diff
@@ -0,0 +1,29 @@
...
+
+# Full URL to the host (and optionally port) where frontend is being served
+FRONTEND_URL=http://localhost:3000
```
mattyg, line 18: It looks like `FRONTEND_URL` is only used in the discord-bot. Should it be moved to `discord-bot/.env`?

mattyg (approved):
I made a few inline notes requesting a few minor changes.

But other than that this is great. Awesome work!!

I was thinking it would be nice to create a separate `docker-compose.development.yml` file for use in a local development environment (with support for live-reload, etc.). But I'll make that another ticket to contain the scope of this.

.env.template.development
kristoferlund: Yes, thats what I thought. Most likely we will need it elsewhere.

packages/api/src/scripts/replace-quantifier.ts
```diff
@@ -67,7 +67,7 @@ interface Arguments {
 }
 
 const run = async (): Promise<void> => {
-  const argv: Arguments = yargs
+  const argv: any = yargs
```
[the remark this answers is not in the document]
kristoferlund: image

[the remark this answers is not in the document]
kristoferlund: Skärmavbild 2022-04-27 kl  19 21 17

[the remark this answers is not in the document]
kristoferlund: Let's make a separate issue about the Arguments.

.env.template.production
kristoferlund: It definitely saves some time when switching from dev to production and vice versa.

mattyg: We shouldn't need to switch from dev to production.

Once you setup a dev or production deployment -- you don't have to touch the env files except extremely rarely.

PR #279, opened 2022-04-12
praise reason markup

nebs-dev:


kristoferlund (approved):
Nice!

There seems to be some diff between discord md and our way of interpreting:

image

image

PR #280, opened 2022-04-12
Feat/229/update activation page

### LINKED ISSUE #229: Give activation page same treatment as login page
These two pages are largely similar. Make sure UX, messages etc are presented similarly in activation page as in login page.

Vyvy-vi:
resolves #229 

This PR makes the Activation page behave like the login page.
*though, I think it would be great if we convert this into a global component


kristoferlund (approved):
👍👍👍

PR #288, opened 2022-04-13
Startup env variables

nebs-dev:


packages/api/.env.template
```diff
@@ -34,3 +34,16 @@ JET_LOGGER_MODE=CONSOLE
 JET_LOGGER_FILEPATH=jet-logger.log
 JET_LOGGER_TIMESTAMP=TRUE
 JET_LOGGER_FORMAT=LINE
+
+
+## Settings ##
+NAME=Praise
```
kristoferlund, line 40: We don't use env file to set these defaults anymore.

kristoferlund (changes requested):

packages/api/.env.template
mattyg: @nebs-dev it looks like these unused ENV values are still in `.env.template` -- can you remove them?

packages/api/src/pre-start/envCheck.ts
```diff
@@ -0,0 +1,17 @@
+import jsonData from './env-required.json';
+
+const envCheck = (): void => {
```
mattyg, line 3: Let's make this function a bit more generic so it can be used by both the discord bot & api.

Can you modify `envCheck()` to take 'jsonData' as a parameter?

packages/api/src/pre-start/env-required.json
```diff
@@ -0,0 +1,19 @@
...
+  "MONGO_PASSWORD",
+  "COOKIE_DOMAIN",
+  "COOKIE_PATH",
+  "SECURE_COOKIE",
+  "JWT_SECRET",
+  "COOKIE_SECRET",
+  "JWT_ACCESS_EXP",
+  "JWT_REFRESH_EXP",
+  "SERVER_URL"
+]
```
mattyg, line 19: - Missing one required env var `NODE_ENV`

Also, curious why you made this file json, rather than just a plain js object? 

Not really that important but I'd say keep it in JS for consistency.

Also, can we add a comment to the top of this file to note that this is where devs should specify required env variables.

mattyg (changes requested):
Hey @nebs-dev. Looks good but I requested some changes for some minor refactoring. LMK if you have questions.

packages/discord-bot/src/pre-start/env-required.json
```diff
@@ -0,0 +1,13 @@
+[
```
mattyg, line 1: - Can we add a comment to the top of this file to note that this is where devs should specify required env variables?

packages/api/src/pre-start/envCheck.ts
nebs-dev: @mattyg do you think we should leave this file in the same place ( packages/api )?

mattyg: yeah for now -- I think we'll want to pull out all the "shared" stuff into a new workspace at some point but we can kick that down the road.

mattyg:
@nebs-dev  I just pushed a few minor changes:
- remove console log
- rename variable for semantic clarity
- refactor a string to use template string syntax rather than the "plus" operand

Otherwise all good!

PR #293, opened 2022-04-14
Bug/use verify quantifier pool size admin

nebs-dev:
- Moved `useVerifyQuantifierPoolSize` from Details to `AssignDialog` as it is used only in that component
- This way it is called only when `AssignDialog` is initialized


mattyg (approved):
:fire: :+1: :1st_place_medal:

PR #306, opened 2022-04-20
Feat/147/whoami command

### LINKED ISSUE #147: `/whoami` Use the praise bot to fetch praise stats, am I activated or not, etc

Vyvy-vi:
Closes #147

This PR adds the `/whoami`  command to the bot, adds an interface for `CommandHandler`s and fixes a typo in `api/src/useraccount/types.ts`("TELEGRAM").
Screenshot 2022-04-21 at 2 08 06 AM
Screenshot 2022-04-21 at 1 31 43 AM


packages/discord-bot/src/commands/whoami.ts
```diff
@@ -0,0 +1,21 @@
+import { SlashCommandBuilder } from '@discordjs/builders';
+import { APIMessage } from 'discord-api-types/v9';
```
mattyg, line 2: looks like this import is unused

packages/discord-bot/src/handlers/whoami.ts
```diff
@@ -0,0 +1,59 @@
...
+  const User = await UserModel.findOne({ _id: ua.user });
+  state.praiseRoles = User?.roles || [];
+  state.address = User?.ethereumAddress || '';
+  state.avatar = ua.avatarId;
+
+  const activatedAccounts = await UserAccountModel.find({ user: ua.user });
+  if (activatedAccounts) {
+    state.activations = [];
+    for (const account of activatedAccounts) {
+      state.activations?.push({
```
mattyg, line 46: I dont think you need the `?` since you just defined `state.activations = [];`

packages/discord-bot/src/utils/stateEmbed.ts
```diff
@@ -0,0 +1,39 @@
+import { MessageEmbed, StringMappedInteractionTypes } from 'discord.js';
```
mattyg, line 1: looks like `StringMappedInteractionTypes` is not used

packages/discord-bot/src/utils/stateEmbed.ts
```diff
@@ -0,0 +1,39 @@
...
+    .setDescription(
+      state.hasPraiseGiverRole && state.activated
+        ? 'Your account is activated and has praise powers.'
+        : state.hasPraiseGiverRole
+        ? "You have praise powers, however your account isn't activated(use `/activate` command)."
+        : "Your account is activated, however you don't have praise powers."
+    )
+    .setThumbnail(
+      `https://cdn.discordapp.com/avatars/${state.id}/${state.avatar || 'oops'}`
+    );
```
mattyg, line 19: why is the ` || oops` necessary? wouldn't state.avatar just be empty?

Also, if we need a conditional - should we make it explicit and only run `setThumbnail` if `state.avatar` is defined?

packages/discord-bot/src/utils/stateEmbed.ts
```diff
@@ -0,0 +1,39 @@
...
+  if (state.activated) {
+    embed.addField('User Roles', state.praiseRoles?.join(' | ') || 'oops');
```
mattyg, line 21: Why is the `|| 'oops'` necessary? Should it be "No roles" or something more specific? Or just blank?

packages/discord-bot/src/utils/stateEmbed.ts
```diff
@@ -0,0 +1,39 @@
...
+    embed.addField('Ethereum Address', state.address || 'oops');
```
mattyg, line 22: Same note as above re `|| 'oops'`

packages/discord-bot/src/utils/stateEmbed.ts
```diff
@@ -0,0 +1,39 @@
...
+      state.activations
+        ?.map(
+          (account) =>
+            `> ${account.platform}\n> Account: ${
+              account.user
+            }\n> Date of Activation: ${formatDate(
+              account.activationDate
+            )}\n> Last Active: ${formatDate(account.latestUsageDate)}`
+        )
+        .join('\n\n') || 'oops'
```
mattyg, line 34: same note as above re `|| 'oops'`

mattyg (changes requested):
I requested a few changes inline. LMK if you have questions. Otherwise looks good!

packages/discord-bot/src/utils/stateEmbed.ts
Vyvy-vi: We could have "No Roles", though the API assigns everyone the "USER" role on the activation Maybe we can change this up a bit and just not create this field if there are no roles, rather than filling it in with text

mattyg (approved):

PR #307, opened 2022-04-20
Remove unused env variables

mattyg:
Remove the following unused env variables from discord-bot.
- `DISCORD_ACTIVATION`
- `DISCORD_MESSAGE`

@Vyvy-vi I did a search for them -- but can you just confirm these env vars are no longer needed?


Vyvy-vi (approved):
yes, they're not needed.

PR #314, opened 2022-04-26
Fix: Only one field updates reactively when submitting a SettingsForm with multiple updated fields 

### LINKED ISSUE #243: Only one field updates reactively when submitting a SettingsForm with multiple updated fields
When submitting a SettingsForm with multiple updated fields, only one update is displayed reactively in the frontend without page refresh.

**Steps to Reproduce**
- Open Settings page
- Change one setting
- Click 'Submit'
*field reflects newly updated value as expected*

- Change several settings
- click 'submit'
*only one of the fields will reflect the updated value as expected (the last one to complete its query)*

nebs-dev:
Changes inside settings model:
-  the way single setting field is saved
- AllSettings is now selector instead of atom
- added AllSettingIds atom

Resolves #243

### DIFF
--- a/packages/frontend/src/model/settings.ts
+++ b/packages/frontend/src/model/settings.ts
@@ -4,14 +4,13 @@ import React from 'react';
 import { toast } from 'react-hot-toast';
 import {
   atom,
+  atomFamily,
   selector,
-  selectorFamily,
   useRecoilCallback,
-  useRecoilState,
   useRecoilValue,
 } from 'recoil';
-import { findIndex } from 'lodash';
-import { ApiAuthGet, useAuthApiQuery } from './api';
+import { ApiAuthGet, useAuthApiQuery, isResponseOk } from './api';
+import { SettingDto } from 'api/dist/settings/types';
 
 export interface Setting {
   _id: string;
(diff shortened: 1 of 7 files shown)


packages/frontend/src/model/settings.ts
```diff
@@ -44,22 +43,71 @@ export const AllSettingsQuery = selector({
...
 
-export const AllSettings = atom<Setting[]>({
+export const SingleSetting = atomFamily<SettingDto | undefined, string>({
+  key: 'SingleSetting',
+  default: undefined,
+});
+
+export const AllSettingIds = atom<string[] | undefined>({
+  key: 'SettingIdList',
+  default: undefined,
```
mattyg, line 53: Should AllSettingIds default to an empty array instead of undefined? Then we don't need to allow the 'undefined' type.

mattyg: actually nevermind i see that's how we're doing it with periods. Let's just keep the same implementations.

mattyg (changes requested):
Looks like you still need to update the local SingleSetting with the received changes from the api. 

Currently the settings changes are not reflected in the recoil state:
- change two settings, click submit
- navigate to another page
- navigate back to the settings page
*changes to settings are not reflected*
- refresh the page
*changes to settings are reflected*

nebs-dev:
> Looks like you still need to update the local SingleSetting with the received changes from the api.
> 
> Currently the settings changes are not reflected in the recoil state:
> 
> * change two settings, click submit
> * navigate to another page
> * navigate back to the settings page
>   _changes to settings are not reflected_
> * refresh the page
>   _changes to settings are reflected_
> 
> Otherwise the changes you made look good!

thanks.. i pushed new changes - updated SingleSettings and now that is fixed
But i introduced another bug, something strange is happening with period settings, i will take a look at that also

packages/frontend/src/model/settings.ts
```diff
@@ -109,22 +141,11 @@ export const useSetSetting = (): useSetSettingReturn => {
...
-          const updatedAllSettings = allSettings.slice();
-          updatedAllSettings.splice(settingIndex, 1, updatedSetting);
-
-          set(AllSettings, updatedAllSettings);
+        if (response.data) {
+          set(SingleSetting(setting._id), response.data);
         }
 
-        toast.success(`Saved setting "${updatedSetting.label}"`);
+        toast.success(`Saved setting "${response.data.label}"`);
```
mattyg: Should this be moved in the conditional above? I.e. only display the success notice if `response.data` is truthy and the local setting is updated?

mattyg (approved):
:fire: :boom: :100:

PR #315, opened 2022-04-26
Refactor/score calculation logic

### LINKED ISSUE #287: Consolidate score calculation logic
Currently the logic to calculate scores is split up / redundantly implemented in the following functions: 
- `exportPraise`
- `praiseWithScore`
- `calculateReceiverScores`
- `calculateQuantificationsCompositeScore`
- `calculateDuplicateScore`
-  `quantificationToDto`.

As this is a critical feature, I'd like to minimize the risk of accidentally introducing bugs. Below are some potential refactoring to reduce the risk:
- move all score calculating logic into the same module for easier review 
- unit tests for all calculation functions

- ~add mongoose 'virtuals' for fetching the calculated score of a praise instance & refactor all score usage to use the above 'virtuals' instead~ Cannot access virtuals from mongoose 'aggregate' queries so this may not be a useful option.
- ~Implement explicit association of quantifications-praise-period with an 'AssignmentModel'~ (moved to Ticket #223)

### LINKED ISSUE #312: Possible incorrect calculation of dismissed scores
Dismissed praise should be included in composite scores as a score of `0`

**From Discord "Praise DEV" group dm:**
> How is the backend handling dismissed praise in regards to the average score right now?  Is it just ignoring it like unfinished one?

> No, I'm almost certain dismissed are counted as 0. Two quantifications, the first = 50, the second = dismissed should yield avg of 25.
> Have to double check..
> Oh. You might have discovered a bug there. And quite an important one 🥶. It does not seem this is handled correctly.

mattyg:
Resolves #287 
Resolves #312 

I noted one question in ticket #287 that I just wanted to flag.

**Overview**
- consolidated score calculation logic into a few functions in `praise/utils/score
- verified all score calculations used in application call the same scoring functions
- Fixed bug where dismissed praise was not included in AVG SCORE calculation for exported csv (#312)
- unit tests for score functions 
- renamed all calculated score values to 'scoreRealized' -- to clearly differentiate from the manually-input 'score'


kristoferlund (approved):
Looks good, great work!

PR #329, opened 2022-05-01
fix(frontend): load period settings when rendering period-specific pages

mattyg:
Fix issue where quantifier slider breaks when loading quantification page directly.

Ensures that per-period settings have been fetched -- on every page that currently uses them.


kristoferlund (approved):
👍👍👍

PR #331, opened 2022-05-01
Feat/prevent push to protected branches

### LINKED ISSUE #316: Restrict pushs directly to 'main' and 'dev' branches
It looks like the github 'branch protection' rules don't seem to prevent pushing directly to protected branches.

We may want to use local git hooks to throw an error when a user tries to push to a protected branch. See https://stackoverflow.com/questions/53726982/setup-pre-push-hook-with-husky-to-prevent-pushing-to-master-branch

mattyg:
Resolves #316 

**Overview**
- add dev dependency `husky`
- add dev dependency `is-git-branch`
- Adds a pre-push hook to throw an error if pushing directly to git branch 'dev' or 'main'
- Noted additional setup step in docs `yarn prepare`


kristoferlund (approved):
👍👍👍

PR #332, opened 2022-05-02
Setup script

kristoferlund:
I created a setup to configure the Praise runtime environment.
- Created a new package `setup`.
- Run: `sh setup.sh` to run script in Docker - eliminates the need to install node, yarn, etc on a production server
- Works both for local development and production

Plus, created a small start script for production environment - `start.sh`. The script won't work until we have merged with main and tagged a new release.


kristoferlund:
> JWT_SECRET and COOKIE_SECRET

Yeah, I was about to post a separate issue for those. Two independent random strings, that's it?

PR #335, opened 2022-05-02
Fix/highlight nav

### LINKED ISSUE #244: Highlight current page in Navs
Currently, the Navs in the sidebar and period pages do not have any visual indication of what page you're on.

Add a visual shading / background color change to indicate the page you're currently on.

Vyvy-vi:
closes #244 

This PR makes the navigation items change colour on selection.

### DIFF
--- a/packages/frontend/src/navigation/NavItem.tsx
+++ b/packages/frontend/src/navigation/NavItem.tsx
@@ -13,8 +13,8 @@ const NavItem = ({ icon, description, to }: NavProps): JSX.Element => {
     <NavLink
       to={to}
       className={(isActive): string =>
-        `relative px-4 py-1 cursor-pointer hover:bg-gray-100 mr-[1px] no-underline flex items-center${
-          isActive ? ' bg-gray-100' : ''
+        `relative px-4 py-1 cursor-pointer mr-[1px] no-underline flex items-center${
+          isActive ? ' bg-gray-800 text-white' : ' hover:bg-gray-100 text-black'
         }`
       }
       id={to.substring(1) + '-nav-button'}
--- a/packages/frontend/src/navigation/NavItem.tsx
+++ b/packages/frontend/src/navigation/NavItem.tsx
@@ -1,6 +1,6 @@
 import { IconProp } from '@fortawesome/fontawesome-svg-core';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { Link } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


mattyg (changes requested):
Looks good!

My only request is can we make the colors different for the active NavItem vs the hovered NavItem?

What about: 
- on hover: bg-gray-100
- on active page: text-white, bg-gray-800

Or something like that

Vyvy-vi:
sounds good to me 👍

kristoferlund:
We need to have a chat about the benefit of highlighting the nav like this. When moving to subpages of main categories the highlight disappears which is.. unexpected behaviour I believe. And since it is difficult to determine for some pages which main "category" they belong to, praise detail page for instance, then it is questionable if highlighting adds any value where it is used.

mattyg:
> We need to have a chat about the benefit of highlighting the nav like this. When moving to subpages of main categories the highlight disappears which is.. unexpected behaviour I believe. And since it is difficult to determine for some pages which main "category" they belong to, praise detail page for instance, then it is questionable if highlighting adds any value where it is used.

I think the only page that can be reached from multiple menu items is the Praise Detail -- otherwise the rest are sub-pages of a menu-item. 

I think you're right: a menu item should be "active" when on any of its sub-pages.

For the Praise Detail page I think it's fine to just not highlight any menu item.

Vyvy-vi:
Q) So should we make an altered version of `NavItem` that doesn't have the highlights, for the PraiseDetail page?

mattyg:
> Q) So should we make an altered version of `NavItem` that doesn't have the highlights, for the PraiseDetail page?

It's already not being highlighted, so no change necessary. My idea was we could display the NavItem in the sidebar as active if any of its "child" pages (in the navigation tree) are active.

But I think we just need to discuss it more before moving forward.

Anyway this is great vy, thank you! We can pick it up while you're on leave this month!

kristoferlund:
> It's already not being highlighted, so no change necessary. My idea was we could display the NavItem in the sidebar as active if any of its "child" pages (in the navigation tree) are active.

Agree. `NavItem` is active if top level page or any of its subpages are active. 

We only need to provider a subtle signal which page the user is on, the dark gray `bg-gray-800` captures attention too much, let's go with `bg-gray-200` instead.

mattyg:
I copied this branch to `fix/highlight-nav` so I could pick it up. So I had to open a new PR #396 -- closing this one.

PR #337, opened 2022-05-03
fix(frontend): center startup loading spinner

mattyg:
Center the startup loading spinner -- it was slightly off


kristoferlund (approved):
👍👍👍

PR #342, opened 2022-05-03
Feat/score rounded to 2 decimal places

### LINKED ISSUE #341: Modify Score calculations to round to 2 decimal places
> Do you think it makes sense to instead round to a few decimal places (say 2)? Do we need to be using integer values? Or should we keep the additional detail? @kristoferlund

@divine-comedian has requested more detail for the duplicate scores. And I agree that the rounding down creates unwanted results. A duplicate of a praise with score 8 gets 0,8 - rounded down to 0. No good. Let's bring decimals back if it is not too much of a hassle. 1 decimal should be enough though.

For the composite score it makes less sense. Your final score is 544,1 or 544. That doesn't make much of a difference.

_Originally posted by @kristoferlund in https://github.com/CommonsBuild/praise/issues/287#issuecomment-1114726417_

mattyg:
Resolves #341

### DIFF
--- a/packages/api/src/praise/utils/score.ts
+++ b/packages/api/src/praise/utils/score.ts
@@ -8,7 +8,7 @@ import { Quantification } from '../types';
 import { getPraisePeriod, isQuantificationCompleted } from './core';
 
 /**
- * Use 2 digits of precision
+ * Digits of precision for rounding calculated scores
  */
 const DIGITS_PRECISION = 2;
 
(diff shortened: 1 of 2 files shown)


kristoferlund (commented):
What would be the motivation for 2 decimals? Don't  1 decimal give us the precision we need?

I just want to keep things as simple as possible.

kristoferlund (approved):
Great! Merging

PR #345, opened 2022-05-03
Fix/reset quantifytable state

### LINKED ISSUE #343: Praise item checkmarks are not cleared when switching pages
When checking a praise at one page and then move on to another page, the check remains which leads to .. unpredictable consequences. :)

image

### LINKED ISSUE #344: Praise details not loading
Praise details seems not to be loading correctly currently.

image

mattyg:
Reset QuantifyTable component state if url params `periodId` or `receiverId` have changed.

Resolves #343


kristoferlund (approved):
Thanks for the quick fix!

PR #349, opened 2022-05-03
Quantifiers displayed as not being done when they in fact are done

### LINKED ISSUE #256: Quantifiers displayed as not being done when they in fact are done
I assume this is a bug of some sort, that it is not a coincidence that three quantifiers all have 299/300 done items.

Skärmavbild 2022-04-07 kl  08 29 18

nebs-dev:
Resolves #256 

- Praise details page:
    - moved quantifications count logic for each quantifier in period from mongo query to utils method
    - used `isQuantificationCompleted()` to count completed quantifications 
    - fix to load quantifier finished items without need to reload the screen

### DIFF
--- a/packages/api/src/period/utils.ts
+++ b/packages/api/src/period/utils.ts
@@ -59,17 +59,14 @@ const quantifiersWithCounts = (
   quantifiers: PeriodDetailsQuantifier[]
 ): PeriodDetailsQuantifierDto[] => {
   const quantifiersWithQuantificationCounts = quantifiers.map((q) => {
-    let finishedCount = 0;
-    q.quantifications.forEach((quantification) => {
-      finishedCount = isQuantificationCompleted(quantification)
-        ? finishedCount + 1
-        : finishedCount;
-    });
+    const finishedQuantifications = q.quantifications.filter((quantification) =>
+      isQuantificationCompleted(quantification)
+    );
 
     return {
       _id: q._id,
       praiseCount: q.praiseCount,
-      finishedCount,
+      finishedCount: finishedQuantifications.length,
     };
   });
 
--- a/packages/api/src/period/utils.ts
+++ b/packages/api/src/period/utils.ts
@@ -59,14 +59,14 @@ const quantifiersWithCounts = (
   quantifiers: PeriodDetailsQuantifier[]
 ): PeriodDetailsQuantifierDto[] => {
   const quantifiersWithQuantificationCounts = quantifiers.map((q) => {
-    const finishedQuantifications = q.quantifications.filter((quantification) =>
+    const finishedCount = q.quantifications.filter((quantification) =>
       isQuantificationCompleted(quantification)
-    );
+    ).length;
 
     return {
       _id: q._id,
       praiseCount: q.praiseCount,
-      finishedCount: finishedQuantifications.length,
+      finishedCount,
     };
   });
 
--- a/packages/api/src/period/utils.ts
+++ b/packages/api/src/period/utils.ts
@@ -14,10 +14,12 @@ import {
   PeriodDocument,
   PeriodDetailsDto,
   PeriodDetailsQuantifierDto,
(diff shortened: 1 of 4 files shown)


packages/api/src/period/utils.ts
```diff
@@ -53,6 +55,27 @@ const receiversWithScores = async (
   return receiversWithQuantificationScores;
 };
 
+const quantifiersWithCounts = (
+  quantifiers: PeriodDetailsQuantifier[]
+): PeriodDetailsQuantifierDto[] => {
+  const quantifiersWithQuantificationCounts = quantifiers.map((q) => {
+    let finishedCount = 0;
+    q.quantifications.forEach((quantification) => {
+      finishedCount = isQuantificationCompleted(quantification)
```
mattyg, line 64: What do you think about refactoring this `forEach` to be a `filter`, and then just count the number of quantifications? with `.length` -- I think it's more readable to avoid for loops when possible.

mattyg (changes requested):
It looks like the period quantifiers page is not displaying the correct finishedCount. See the screencast below:

https://user-images.githubusercontent.com/159270/166609794-390e015e-294b-4a7a-af7c-c00d546b85fc.mp4

packages/api/src/period/utils.ts
nebs-dev: actually i first did it that way and than changed to this because it should be faster but i think it will not impact user experience either way... ok, will revert my changes

packages/api/src/period/utils.ts
```diff
@@ -59,17 +59,14 @@ const quantifiersWithCounts = (
...
+    const finishedQuantifications = q.quantifications.filter((quantification) =>
+      isQuantificationCompleted(quantification)
+    );
 
     return {
       _id: q._id,
       praiseCount: q.praiseCount,
-      finishedCount,
+      finishedCount: finishedQuantifications.length,
     };
```
mattyg, line 70: i don't mean to be a stickler but what do you think of this code style / implementation? Just for a bit cleaner / more readable code:

```
const finishedCount = q.quantifications.filter((quantification) =>
      isQuantificationCompleted(quantification)
    ).length;

    return {
      _id: q._id,
      praiseCount: q.praiseCount,
      finishedCount,
};
```

packages/frontend/src/pages/Periods/Details/PeriodDetailsPage.tsx
```diff
@@ -75,6 +75,8 @@ const PeriodDetailPage = (): JSX.Element => {
   const [detailsLoaded, setDetailsLoaded] = React.useState<boolean>(false);
   const { path, url } = useRouteMatch();
 
+  PeriodDetailLoader();
+
```
mattyg, line 79: Lets just delete the PeriodDetailsLoader entirely, and just move the 2 relevant lines into the PeriodDetailPage component:
```
  const { location } = useHistory();
  useSinglePeriodQuery(periodId, location.key);
```

- then we can return `null` if the necessary data isn't loaded:
```
  if (!detailsLoaded || !period || !period.recievers) return null;
```

- then we can delete the useEffect entirely

mattyg (changes requested):
requested one more inline refactor above. Almost there!!

packages/frontend/src/pages/Periods/Details/PeriodDetailsPage.tsx
nebs-dev: yeah, i wanted to do that but problem is that i am getting this typescript error:
image

mattyg: Just add null as a return type for the component:

```
const PeriodDetailPage = (): JSX.Element | null => {
```

mattyg (approved):
The function `PeriodDetailsLoader` is now unused and can be deleted. Otherwise all good!

Issue #80, opened 2022-02-17
Opening add quantifier dialog should place focus on input


kristoferlund:
As well as dialog to mark praise as duplicate.

![Skärmavbild 2022-02-17 kl. 11.02.50.png](https://images.zenhubusercontent.com/5fdba670baa205320c5a291e/745643da-71fe-4cf7-9eb3-3449a4915ce1)

PR #356, opened 2022-05-04
Improve dapp responsiviness

### LINKED ISSUE #190: Improve page content responsiveness
**Overview**
- Refactor QuantifyTable component to support small viewports without scroll overflow.
- Consider increasing max width of most page content areas
- Consider picking a standard  max width for *all* page content areas

pkretzschmar:
Improve app responsiviness specially for mobile devides. Fix #190


packages/frontend/src/pages/Users/components/AddDialog.tsx
```diff
@@ -6,7 +6,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { Dialog } from '@headlessui/react';
 import { UserDto, UserRole } from 'api/dist/user/types';
 import { useCombobox } from 'downshift';
-import React from 'react';
+import React, { LegacyRef, useEffect, useRef } from 'react';
```
mattyg, line 9: looks like we're not using `LegacyRef` import

mattyg (changes requested):
This is awesome! Thank you so much for fixing #80 too!

I think there's a few more changes in #190 we might want to think about:

- What do you think about picking a standard max width for every page? Currently there are some with `max-w-2xl`, `max-w-4xl`, and `max-w-5xl`. I feel like this causes a bit weird, jumpy UX when the user navigates to a different page, and the content area jumps to a different width. But it's also a UX pattern you see in a lot of sites. Any thoughts on it? I'm really not sure what the best approach is.

- At the very least, if we don't want to increase the max width of *all* pages, lets increase it for the `PeriodDetailsPage` which has a lot of content and feels a bit "tight".

- refactoring the QuantifyTable component from a table to divs, so that it can be made more responsive in smaller screens. On a smaller screen, the columns could be split into two rows, with a bit more spacing between each list item. Something like this wireframe: https://wireframe.cc/SVaaNp -- do you think that would be preferable?

kristoferlund (approved):
Great @pkretzschmar! I'm merging this now, let's revisit the topic when we switch to top navigation. We might need to make additional changes then.

PR #362, opened 2022-05-05
Fix/activate useraccount only once

### LINKED ISSUE #357: A user should only be allowed to activate their account once

mattyg:
Resolves #357 

Responds with an error message if the user tries to activate their account via the discord bot, and it has already been activated.


kristoferlund (approved):
Nice!

Vyvy-vi:
@mattyg How would people change the User associated with their discord account?

PR #364, opened 2022-05-05
Feat/ Add giver and receiver id to exported praise csv

### LINKED ISSUE #359: Add giver & receiver useraccount._id to exported csvs
The RAD dashboard has some charts to track a person's engagement with the praise system over time. 

However, the only data included in the exported csv to identify the user's is their discord name, which can be changed at any time.

Ideally we would include the `user._id` in the exported csv, to track the human behind the discord account, but currently anyone can create a user without having an associated discord account.

Since all praise given are between useraccounts, it makes sense to just include the useraccount._id for the giver and receiver in the exported csv.

mattyg:
Resolves #359

### DIFF
--- a/packages/api/src/period/controllers.ts
+++ b/packages/api/src/period/controllers.ts
@@ -604,12 +604,20 @@ export const exportPraise = async (
       value: 'receiver.name',
     },
     {
+      label: 'TO USER ACCOUNT ID',
+      value: 'receiver._id',
+    },
+    {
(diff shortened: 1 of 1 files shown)


0xNuggan:
Thanks! Seems good to me. We'd want to reexport all past rounds again, though, to not have changing data structures.

0xNuggan (approved):

PR #366, opened 2022-05-06
Feat/156/sanitise praise reason

### LINKED ISSUE #156: Sanitise reason content string returned by discord
Discord return's channel mentions, user mentions, emotes and role mentions as formatted snowflakes inside the content string.
This breaks the context in people's praise.

Vyvy-vi:
resolves #156 

This PR adds a new optional field to the praise schema: `reasonRealized`, which is a normalised string containing plain text instead of discord object id formating.

Screenshot 2022-05-06 at 1 36 56 PM


packages/api/src/praise/entities.ts
```diff
@@ -31,6 +31,7 @@ const QuantificationModel = mongoose.model<QuantificationDocument>(
 const praiseSchema = new mongoose.Schema(
   {
     reason: { type: String, required: true },
+    reasonRealized: { type: String, required: false },
     sourceId: { type: String, required: true },
```
mattyg, line 35: I think `reasonRealized` should be required for all Praise entities. 

Then we can remove all related the conditionals all the way down through the frontend.

We'll need to write a migration to update all the praise in the database so they also have a `reasonRealized` value.

mattyg (changes requested):
Thanks vy!

I think `reasonRealized` should be required for all Praise entities. We'll need to write a migration to update all the past Praise with a `reasonRealized` value.

Then we can remove all related the conditionals all the way down through the frontend.

I might recommend just removing the `reason` parameter from the 'PraiseDto' entirely -- so they only data we have available in the frontend is the `reasonRealized`.

I know you're on leave right now, so we can pick this up while your gone, unless you wanted to work on it! Just let us know

mattyg:
This work is continued a new PR #397. Closing

PR #367, opened 2022-05-06
Bug/discord markup

### LINKED ISSUE #290: Fix diff between discord markdown and our parsing
There seems to be some diff between discord md and our way of interpreting:

image

image

nebs-dev:
Resolves #290

- removed https://www.npmjs.com/package/micromark as i couldn't make it work with discord markup, none of its extensions do what we need it seems
- tried https://www.npmjs.com/package/discord-markdown - works but it looks like it is not mantained
- created small markup script that uses regex


mattyg (changes requested):
:fire: Great!

Do you think we should add these formatting styles as well? I'm sure they'd be used rarely but might make sense to just include them anyway:
- \`inline code\` -> `<code>inline code</code>`
- ` ```multi line code``` `-> `<pre><code>multi line code</code></pre>`

And also it looks like the formatting is not being applied in the following pages, because they're not using the Praise component:
- QuantifyTable
- ReceiverSummaryTable

I think we'll just want to refactor these pages to use the standard Praise component so the formatting is applied everywhere.

nebs-dev:
> fire Great!
> 
> Do you think we should add these formatting styles as well? I'm sure they'd be used rarely but might make sense to just include them anyway:
> 
> * `inline code` -> `<code>inline code</code>`
> * `` ```multi line code``` ``-> `<pre><code>multi line code</code></pre>`
> 
> And also it looks like the formatting is not being applied in the following pages, because they're not using the Praise component:
> 
> * QuantifyTable
> * ReceiverSummaryTable
> 
> I think we'll just want to refactor these pages to use the standard Praise component so the formatting is applied everywhere. Though the Praise component might need some refactoring to handle the differences with the QuantifyTable version.

yup, i will add formatting for the code also.. i forgot about those because i was checking this link:
https://support.discord.com/hc/en-us/articles/210298617-Markdown-Text-101-Chat-Formatting-Bold-Italic-Underline-

about other pages, it is like i vaguely remember that we for some reason decided to skip formatting there but i am not 100% sure.. @kristoferlund ?

kristoferlund:
Hm. Don't remember either why we would not format `code` in some places. Format everywhere to start and let's see how that looks.

nebs-dev:
> fire Great!
> 
> Do you think we should add these formatting styles as well? I'm sure they'd be used rarely but might make sense to just include them anyway:
> 
> * `inline code` -> `<code>inline code</code>`
> * `` ```multi line code``` ``-> `<pre><code>multi line code</code></pre>`
> 
> And also it looks like the formatting is not being applied in the following pages, because they're not using the Praise component:
> 
> * QuantifyTable
> * ReceiverSummaryTable
> 
> I think we'll just want to refactor these pages to use the standard Praise component so the formatting is applied everywhere. Though the Praise component might need some refactoring to handle the differences with the QuantifyTable version.

In the end i had to refactor <Praise> component and also with this solution, i had to add library to linkify praise reason as we were missing that also.
I moved a lot of methods from `QuantifyTable` to `Praise` as we need it there to show praise correctly ( dismissed, disabled, etc )

mattyg:
> In the end i had to refactor component and also with this solution, i had to add library to linkify praise reason as we were missing that also. I moved a lot of methods from `QuantifyTable` to `Praise` as we need it there to show praise correctly ( dismissed, disabled, etc )

Let's go over this together early next week. I think we can refactor the Praise component in a much cleaner way without needing to conditionally include a bunch of logic that will only be used in certain contexts.

kristoferlund:
> Basically I think it should just be a container for the "styling" and common elements of the Praise component.

While you are looking at the Praise component, could you also include props to show/hide other details? 
- Score
- Praise Id

Details can for now be displayed below praise text like this:

image

But we also need to remake the layout of the component soon.

mattyg (changes requested):
Looks like we still need to refactor `ReceiverSummaryTable` to use the new Praise component -- otherwise looks great!

mattyg (approved):
made a few minor changes:
- fixed build issues in api & discord-bot (likely resulting from all the dependency version bumps)
- add space before "to" in Praise component
- refactor Praise component to include InlinePraiseButtons internally (to ensure that we can strikethrough / mute the reason text when praise is dismissed / marked duplicate)

PR #368, opened 2022-05-09
Fix/server url fix

### LINKED ISSUE #348: Revert overwriting image url settings by environment
I missed this change in my review:  https://github.com/CommonsBuild/praise/pull/340/commits/eda9fcecaa7aac1e4e13b6aaf71d3c28045dda26

Do we *need* to overwrite the specified env variable? 
- The env variable SERVER_URL can simply include a port or not.
- I.e. in the development environment just include a port in the SERVER_URL env variable.
- We can just modify  `setup/src/index.ts` script to simply include a port in SERVER_URL for development enviornments

If you think we absolutely *must* overwrite the specified env variable -- Can we please consolidate it in one place -- on a high-level module like `server.ts`?  The per-environment overwrites can very easily become scattered around the codebase and become a nightmare to make sense of / maintain.

Edit: also it looks like this broke the logo rendering on production.

kristoferlund:
Resolves #348

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -11,7 +11,8 @@ HOST=localhost
 # on $HOST/api
 API_PORT=8088
 
-# Full URL to the host (and optionally port) where the API is running
+# Full URL to the host where the API is running.
+# When running in development, the URL should also include API_PORT
 SERVER_URL=https://localhost
 
 # Full URL to the host (and optionally port) where frontend is being served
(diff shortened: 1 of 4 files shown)


mattyg (approved):
Code changes look good, thank you!

I confirmed images display properly on a production deployment. But I can't seem to get development or local docker deployment to work at all (502 responses from api). I think that's probably a separate / previously existing issue though.

PR #369, opened 2022-05-09
Setup script now cleans up any dangling Docker images

kristoferlund:


mattyg (changes requested):
Do you think the setup script should be responsible for cleaning up a sysadmin's docker data?

I'd be a little worried that we could delete some other images on the same server, not related to praise, that the sysadmin cares about.

If you want to keep it, I think we could use the `--filter` flag to only delete *our* docker images?

Also, maybe we should add a "confirmation" to the setup.sh script that says something like: "THIS WILL DELETE ALL UNUSED DOCKER IMAGES, ARE YOU SURE YOU WANT TO CONTINUE? (y/n)" -- kind of like the warning message displayed when you run `docker image prune`

kristoferlund:
You are right, perhaps the setup script is not the right place for a cleanup. Closing.

PR #372, opened 2022-05-10
FAQ page

### LINKED ISSUE #219: add FAQ to sidebar in praise dashboard
We get a lot of 1 off questions - it would be good to identify and answer some of the most common questions from quantifiers right in the UI without making them scan through 3-4 forum posts looking for answers. 

image

I suggest we can make a simple JSON file with questions and answers in them - we could toggle the FAQ off or let new orgs build their own JSON that populates the FAQ page - here is an example from https://giveth.io/faq

image

nebs-dev:
Resolves #219 

Added:

- new settings migration
- settings validator for FAQ json
- FAQ route and page that shows data from settings table


packages/frontend/src/pages/FAQ/FAQPage.tsx
```diff
@@ -0,0 +1,61 @@
...
+          {faq.map((item, index) => {
+            return (
+              <Disclosure
+                defaultOpen={index === 0 ? true : false}
+                as="div"
+                key={index}
+              >
+                {({ open }) => (
+                  <>
+                    <Disclosure.Button className="flex w-full justify-between rounded-lg bg-gray-200 px-4 py-2 text-left text-sm font-medium text-gray-900 hover:bg-gray-400 focus:outline-none focus-visible:ring focus-visible:ring-purple-500 focus-visible:ring-opacity-75">
```
kristoferlund, line 37: Let's simplify the styling a bit:
```
flex justify-between w-full px-4 py-2 mb-2 font-medium text-left bg-gray-100 rounded-lg hover:bg-gray-200 focus:outline-none
```

packages/frontend/src/pages/FAQ/FAQPage.tsx
```diff
@@ -0,0 +1,61 @@
...
+                      <span>{item.question}</span>
+                      <FontAwesomeIcon
+                        icon={faCaretUp}
+                        size="1x"
+                        className={`${
+                          open ? 'rotate-180 transform' : ''
+                        } h-5 w-5 text-gray-500`}
```
kristoferlund, line 44: Same here, simplify, remove `text-gray-500`.

packages/frontend/src/pages/FAQ/FAQPage.tsx
```diff
@@ -0,0 +1,61 @@
...
+                      />
+                    </Disclosure.Button>
+                    <Disclosure.Panel className="px-4 pt-4 pb-2 text-sm text-gray-500">
```
kristoferlund, line 47: Simplify:
```
px-4 pt-1 pb-3
```

packages/frontend/src/navigation/Nav.tsx
```diff
@@ -8,6 +8,7 @@ import {
   faCalculator,
   faCog,
   faPrayingHands,
+  faQuestion,
```
kristoferlund, line 11: Go with `faQuestionCircle` instead.

packages/api/src/settings/validators.ts
```diff
@@ -13,6 +13,22 @@ export function fieldTypeValidator(
     return isNumeric(this.value);
   }
 
+  if (this.type === 'Textarea' && this.key === 'PRAISE_FAQ') {
```
kristoferlund, line 16: Yes, we want to validate the input. But, is this the place? Perhaps it would be ok to do a general "Valid JSON" check and introduce JSON as a new settings type? What do you say @mattyg?

kristoferlund:
Overall, looks good! I made some styling comments to make look more inline with the rest of the current styling. And we should consider if we need some more generalised way to handle settings in JSON format.

kristoferlund (changes requested):
Overall, looks good! I made some styling comments to make look more inline with the rest of the current styling. And we should consider if we need some more generalised way to handle settings in JSON format.

kristoferlund:
image

kristoferlund:
Changed base from `main` to `dev` also. Let's continue merging to dev until we decide otherwise.

packages/api/src/settings/validators.ts
mattyg: Hmm. I agree there should be a new setting type. But let's make it specifically for a JSON file that formatted as a "question-answer" list, say  "QuestionAnswerJSON".

I think it makes sense to validate both that it's valid JSON, and that the valid JSON has the expected data fields. We could use json-schema to define the expected schema and validate it.

I agree it feels a bit weird to validate the "question-answer" format here, but I'm not sure if there is a better place, since the data is input as a json text via the settings page.

mattyg: Honestly it seems kinda weird in general to have this data be input as a json file. Do we expect end-users will have an easy time editing json documents? They're kinda particular and require very specific formatting (like no trailing commas, etc.). Or do we expect to include a default FAQ page that will likely not be edited (or only very rarely) by end-users?

If we expect they'll want to do a lot of customization of the FAQ, would it make more sense to have a UI widget for manually adding each individual question and answer? Then setting up another set of api entities, controllers, and routes for adding, removing, and editing questions and answers? We could include markdown formatting for answers too. 

Perhaps that should be stage 2 of this feature, to keep the scope contained.

Vyvy-vi: I think markdown would be better.

Vyvy-vi: We could outsource this, and just ask users to add a link to a HackMD or a github gist or a markdown file

nebs-dev: What we decided about this? Mitch said in the meeting that he is fine with json file format..
So should i create a new setting type and validate that input is valid JSON and also that it is in question-answer format?

mattyg: > What we decided about this? Mitch said in the meeting that he is fine with json file format.. So should i create a new setting type and validate that input is valid JSON and also that it is in question-answer format?

Yes, I think that's fine for v1 of this feature. Down the road we can consider implementing markup and a better UX for changing the contents.

For this PR yeah lets just do exactly what you said: create a new setting type which will validate that the input is JSON and in the question-answer format.

kristoferlund (changes requested):

packages/api/src/database/migrations/07_settings_faq.ts
```diff
@@ -0,0 +1,32 @@
+import { SettingsModel } from '../../settings/entities';
+
+const settings = [
```
kristoferlund, line 3: Setting need to be added to a `SettingGroup`, introduced in #399.

kristoferlund (changes requested):
Looks good but was unable to run. See comment about `SettingGroup`.

- `07_settings_faq.ts` should be named `09_settings_faq.ts`

kristoferlund (approved):

PR #395, opened 2022-05-12
Create user screen

### LINKED ISSUE #320: Convert the Quantifier Pool screen to a User Admin screen
I made wireframes for the user admin screen. 
- Filter and search users on the first screen
- Set roles on the detail screen

.. anything else right now?

image

Miro board: https://miro.com/app/board/o9J_lyTkWjY=/

pkretzschmar:
Resolves #320


packages/frontend/src/navigation/AuthenticatedRoutes.tsx
```diff
@@ -78,10 +81,14 @@ const Routes = (): JSX.Element => {
...
-      <AuthRoute roles={[ROLE_ADMIN]} path={'/pool'}>
+      <AuthRoute roles={[ROLE_ADMIN]} exact path={'/pool'}>
         <UsersPage />
       </AuthRoute>
 
+      <AuthRoute roles={[ROLE_ADMIN]} path={'/pool/:userId'}>
+        <UserDetailsPage />
+      </AuthRoute>
+
       <Route exact path={'/periods'}>
```
mattyg, line 92: Let's change the URL for the user page from `/pool/*` to `/users/` to reflect that its now a general-purpose user-management page, not just for the quantifier users.

packages/frontend/src/pages/UserDetails/UserDetailsPage.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const UserDetailsPage = (): JSX.Element => {
+  const { userId } = useParams<SingleUserParams>();
+  const user = useRecoilValue(SingleUser({ userId }));
+  const { addRole, removeRole } = useAdminUsers();
+
+  const handleRole = (role: UserRole, user: UserDto): void => {
+    let resp;
+    const isRemove = user.roles.includes(role);
+    async (): Promise<void> => {
```
mattyg, line 22: It looks like you declare this function but it's never run? It doesn't seem to make any network requests when 'handleRole' is called.

Why not just modify `handleRole` to return a `Promise<void>`? Then you can get rid of this nested function.

packages/frontend/src/pages/UserDetails/UserDetailsPage.tsx
```diff
@@ -0,0 +1,88 @@
...
+    };
+  };
+
+  return (
+    <div className="max-w-2xl mx-auto">
+      <BreadCrumb name="User details" icon={faUserGroup} />
+      <BackLink to="/pool" />
+      <div className="praise-box flex flex-col gap-2">
+        <span>User identity</span>
+        {!!user && (
```
mattyg, line 40: the `!!` is not necessary -- if user is defined it is "truthy"

packages/frontend/src/pages/UserDetails/UserDetailsPage.tsx
```diff
@@ -0,0 +1,88 @@
...
+            <span>Created: {formatIsoDateUTC(user.createdAt)}</span>
+            <span>Last updated: {formatIsoDateUTC(user.updatedAt)}</span>
+          </>
+        )}
+      </div>
+      <div className="praise-box flex flex-col gap-2">
+        <span>Linked Discord identities</span>
+        {user?.accounts?.map((account, index) => (
+          <>
+            <span>User account #{index + 1}</span>
```
mattyg, line 54: This text "User Account #x" is not necessary.

packages/frontend/src/pages/UserDetails/UserDetailsPage.tsx
```diff
@@ -0,0 +1,88 @@
...
+            <span className="text-xl font-bold">{account.name}</span>
+            <span>Discord User ID: {account.user}</span>
+            <span>Created: {formatIsoDateUTC(account.createdAt)}</span>
+            <span>Last updated: {formatIsoDateUTC(account.updatedAt)}</span>
```
mattyg, line 58: Can we tighten the spacing of the last 3 lines slightly? To better reflect the mockup.

packages/frontend/src/pages/UserDetails/UserDetailsPage.tsx
```diff
@@ -0,0 +1,88 @@
...
+              >
+                <input
+                  checked={user.roles.includes(role)}
+                  className="text-lime-500 cursor-pointer"
+                  name={role}
+                  type="checkbox"
+                  onChange={(): void => handleRole(role, user)}
+                />
+                <span className="text-white">{role}</span>
+              </div>
```
mattyg, line 80: Unchecked roles should have a gray backround:

image

packages/frontend/src/pages/UserDetails/UserDetailsPage.tsx
```diff
@@ -0,0 +1,88 @@
...
+                key={role}
+                className="flex gap-2 justify-center items-center bg-black py-2 px-3 rounded-md cursor-pointer"
+                onClick={(): void => handleRole(role, user)}
+              >
+                <input
+                  checked={user.roles.includes(role)}
+                  className="text-lime-500 cursor-pointer"
+                  name={role}
+                  type="checkbox"
+                  onChange={(): void => handleRole(role, user)}
```
mattyg, line 77: You don't need the `onChange` on the input element, because you already have an `onClick` on it's parent div.

(right now if you click the checkbox, it calls handleRole twice -- triggered once by input onChange, and again by div onClick).

packages/frontend/src/pages/Users/components/UsersTableRow.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const history = useHistory();
+  return (
+    <div
+      className="my-3 px-4 flex justify-between items-center cursor-pointer h-12 rounded-md hover:bg-gray-100"
+      onClick={(): void => history.push(`pool/${data._id}`)}
+    >
+      <div className="w-1/3 flex items-center gap-4">
+        <UserAvatar user={data} />
+        <span className=" font-mono text-sm">
+          {shortenEthAddress(data.ethereumAddress!)}
```
mattyg, line 22: The user.ethereumAddress should always be defined (according to the mongoose entity). 

Thus I think we can should modify the User & UserDto types to reflect this (make ethereumAddresss no longer optional). Then you won't need this non-null assertion.

packages/frontend/src/pages/Users/components/UsersTableRow.tsx
```diff
@@ -0,0 +1,37 @@
...
+        </span>
+      </div>
+      <div className="w-1/3">{getUsername(data)}</div>
+      <div className="w-1/3">
+        {data.roles.map((role, index) => {
+          if (role !== UserRole.USER) {
+            return <RoleBadge key={`${role}-${index}`} label={role} />;
```
mattyg, line 29: I think we can use the existing component `InlineLabel` for these role badges. (good to keep re-using standard components to keep styling consistent, and also they're a bit smaller which reflects the mockup).

packages/frontend/src/pages/Users/components/UsersTablePagination.tsx
```diff
@@ -0,0 +1,62 @@
...
+  setPage: React.Dispatch<React.SetStateAction<number>>;
+}
+
+const UsersTablePagination = ({
+  lastPage,
+  page,
+  setPage,
+}: IUsersTablePagination): JSX.Element => {
+  return (
+    <div className="flex gap-2 items-center select-none mt-8 pl-4">
```
mattyg, line 22: Let's increase the spacing between these arrow elements slightly to reflect the mockup:

image

packages/frontend/src/pages/Users/components/UsersTableHeader.tsx
```diff
@@ -0,0 +1,13 @@
+interface IUsersTableHeader {
+  label: string;
+}
+
+const UsersTableHeader = ({ label }: IUsersTableHeader): JSX.Element => {
+  return (
```
mattyg, line 6: I don't think it's necessary to make this it's own component

packages/frontend/src/pages/Users/components/UsersTable.tsx
```diff
@@ -1,121 +1,159 @@
...
-      },
-    ],
-    []
+  const allUsers = useRecoilValue(AllUsers);
+  const [tableData, setTableData] = React.useState<UserDto[]>();
+  const [selectedRole, setSelectedRole] = React.useState<UserRole>(
+    UserRole.USER
   );
+  const [filter, setFilter] = React.useState<string>('');
+  const [filterType, setFilterType] = React.useState<string>('user');
```
mattyg, line 34: I think we can remove the `filterType` and conditional entirely.

Instead, always search *both* the username, and ethereumAddresses.

packages/frontend/src/pages/Users/components/UsersTable.tsx
```diff
@@ -1,121 +1,159 @@
...
+          if (username?.includes(filter.toLocaleLowerCase())) {
+            return user;
+          }
+        });
+      } else {
+        filteredData = tableData.filter((user) => {
+          if (
+            user.ethereumAddress?.toLocaleLowerCase() ===
+            filter.toLocaleLowerCase()
+          ) {
```
mattyg, line 59: The ethereumAddress search doesn't seem to be working. I can't get any results to appear.

Also, perhaps it should also check if the ethereumAddress *includes* the search text, instead of *equaling* the search text (just like you're doing with the user name search above).

packages/frontend/src/pages/Users/components/UsersTable.tsx
```diff
@@ -1,121 +1,159 @@
...
+            return user;
+          }
+        });
+      }
+      setTableData(filteredData);
+    }
+  }, [filterType, filter]);
```
mattyg, line 66: getting a lint warning: useEffect missing dependency 'tableData'

packages/frontend/src/pages/Users/components/UsersTable.tsx
```diff
@@ -1,121 +1,159 @@
...
+        break;
+      case UserRole.FORWARDER:
+        setTableData(allForwarderUsers);
+        break;
+      case UserRole.QUANTIFIER:
+        setTableData(allQuantifierUsers);
+        break;
+    }
+    setFilter('');
+  }, [selectedRole]);
```
mattyg, line 84: getting a lint warning: `React Hook React.useEffect has missing dependencies: 'allAdminUsers', 'allForwarderUsers', 'allQuantifierUsers', and 'allUsers'. Either include them or remove the dependency array`

packages/frontend/src/pages/Users/components/UsersTable.tsx
```diff
@@ -1,121 +1,159 @@
...
+              size="1x"
+              className="absolute top-1/2 transform -translate-y-1/2 left-3"
+            />
+            <input
+              type="text"
+              name="search"
+              placeholder="Search"
+              className="bg-transparent outline-none border-none focus:ring-0 block pl-8"
+              value={filter}
+              onChange={(event: React.ChangeEvent<HTMLInputElement>): void =>
```
mattyg, line 126: The search results update when I *add* characters to the search text, but not when I *remove* characters.

It also needs to reset to showing all results when the search query is empty.

mattyg (changes requested):
Thanks @pkretzschmar! This is great!!

I left a bunch of inline comments with requested changes (and a few things that weren't working). 

Let me know if you want to go over any of this together! Or even pair-program some of it if you're interested :smiley:

packages/frontend/src/navigation/AuthenticatedRoutes.tsx
kristoferlund: Agree

packages/frontend/src/pages/UserDetails/UserDetailsPage.tsx
kristoferlund: Yes, in general, always make checks for required data as early as possible and return `null` if something is missing, rather than returning partially complete components.

packages/frontend/src/pages/UserDetails/UserDetailsPage.tsx
kristoferlund: Yes, a user can only have one linked Discord and/or Telegram account. Unless you are running locally for development. This happens when you start with a fresh db with dummy data and then activate your account using the bot. Two Discord user accounts, should not be possible right @mattyg. I created a new issue #398.

packages/frontend/src/pages/UserDetails/UserDetailsPage.tsx
kristoferlund: Mockup in Miro doesn't allow for any custom line spacing. Avoid `<span>` when not explicitly needed. In this case you can get default line height by wrapping in a `<div>` instead and add linebreaks `<br>`.

packages/frontend/src/pages/Users/components/UsersTableRow.tsx
kristoferlund: Doesn't this collide with the needs of #361? If the User is created before account activation we don't know the eth address.

kristoferlund:
Yes, good stuff @pkretzschmar!

A little bit of polishing needed, then it's done.

packages/frontend/src/pages/Users/components/UsersTableRow.tsx
mattyg: > The user.ethereumAddress should always be defined (according to the mongoose entity).

That is outside the scope of this PR. When we're ready to address #361 there will need to be changes throughout the codebase.

packages/frontend/src/pages/UserDetails/UserDetailsPage.tsx
mattyg: > Yes, a user can only have one linked Discord and/or Telegram account. Unless you are running locally for development. This happens when you start with a fresh db with dummy data and then activate your account using the bot. Two Discord user accounts, should not be possible right @mattyg. I created a new issue #398.

To be clear though -- this is not an issue with this PR, and requires no changes.

packages/frontend/src/pages/UserDetails/UserDetailsPage.tsx
pkretzschmar: The ESlint pattern that we are currently using doesn't allow us to use a function that returns Promise<void> on the onClick properties, because they're expecting a void return. I can just disable [the rule](https://typescript-eslint.io/rules/no-misused-promises/) for the file, but I really don't know if this is a good idea.

packages/frontend/src/pages/Users/components/UsersTableRow.tsx
pkretzschmar: I changed to the component that you said. It's a bit different from what I saw on the design, but I agree with you about re-using stuff.

packages/frontend/src/pages/UserDetails/UserDetailsPage.tsx
mattyg: You can actually put a `void` in front of the function call and typescript will ignore the return type.

So this: `onClick={async (): Promise<void> => handleRole(role, user)}` would change to: `onClick={(): void => void handleRole(role, user)}`

mattyg (changes requested):
Awesome! :fire:

I went through and made a bunch of minor changes:
- deleted two components that are no longer needed
- changed label of users page NavItem to "Users"
- remove eslint ignore unused promises (see inline message response)
- remove new dependency 'classnames' -- we've been using a local utility 'classNames' in `utils/index.ts` for the same functionality
- hide ethereum address when it is identical to user name
- remove frontend function `getUsername` -- instead use `user.nameRealized` (reflect refactoring done in #365 while this PR was open) 
- remove frontend implementation of shortenEthAddress, use implementation (reflect refactoring done in #365 while this PR was open) 
- prevent list of roles from breaking mid-word


I think there are just 3 issues remaining. I made screencasts to demonstrate them:
- The pagination buttons do not change when the users list is limited to searched text:

https://user-images.githubusercontent.com/159270/169884071-371b95f8-3cb5-4634-baf8-4cb463c9984a.mp4



- The UserTable overlaps itself on small screens, and overflows the content area.
Screenshot_2022-05-23_14-07-42


- The role toggle buttons seem to be able to get into a weird state where they don't reflect the backend user.roles. I'm pretty sure this is related to use of `useRecoilCallback` within `useAdminUsers`. The frontend state setup is kinda confusing, so please let me know if I can help! We could also pair-program through this sometime if that's helpful!

https://user-images.githubusercontent.com/159270/169883996-de218f3a-d4b9-4f90-a92b-32572384918d.mp4

mattyg:
Also @pkretzschmar I know you've got other projects you're working on -- if you have limited availability to finish up this PR and #356 just let me know and I can pick up where you left off!

kristoferlund (approved):
Yes! Nice work @pkretzschmar! I tweaked some paddings and made some other smaller fixes. Merging this now.

PR #397, opened 2022-05-15
Feat/sanitise praise reason

### LINKED ISSUE #156: Sanitise reason content string returned by discord

mattyg:
Resolves #156 

**Deployment Steps**
- Add `DISCORD_TOKEN` env variable to `api/.env`
- Add `DISCORD_GUILD_ID` env variable to `api/.env`
- install dependencies
- Run migrations (will run on api launch)

**Overview**
- Copy work from #366 
- Add dependency `discord.js` to api
- helper utility to generate `reasonRealized` from given discord client, sourceId, and reason
- Migration to generate `reasonRealized` for existing praise documents without it
- Generate `reasonRealized` for imported praise within import-praise command
- Make `reasonRealized` a required field in Praise model, types -- remove `reason` field from PraiseDto
- Remove all frontend conditionals which check if `reasonRealized` is defined


packages/api/src/praise/utils/core.ts
```diff
@@ -74,3 +75,37 @@ export const isQuantificationCompleted = (
...
+
+  return discordClient;
+};
+
+export const generateReasonRealized = async (
+  discordClient: Client,
+  sourceId: string,
+  reason: string
+): Promise<string> => {
+  const parsedSourceId = sourceId.match(/DISCORD:[\d]+:([\d]+)/);
```
kristoferlund, line 98: When running a local development setup, API startup gets stuck here because of invalid Discord source Id.

kristoferlund (changes requested):
Looks good! We need some workaround to be able to run locally with seeded database data though.

Plus, I made a small addition to the setup script to ensure Discord ENV variables end up in the .env file.

kristoferlund (approved):
Great!

PR #404, opened 2022-05-24
Display Praise dates as relative, with absolute date on hover

### LINKED ISSUE #249: Display Praise dates as relative, with absolute date on hover.
from kristofer: https://github.com/CommonsBuild/praise/issues/204#issuecomment-1089081644

The Discord way of presenting the dates is nice, with a short relative date written out and the full date on hover.

image

All praise dates should be displayed in users local tz.

nebs-dev:
Resolves #249 

- modified date for Praise component
- calculate difference in time from now to show instead of date ( if today or yesterday )
- use react tooltip to display full date on hover


packages/frontend/src/utils/date.ts
```diff
@@ -5,6 +5,7 @@ import jstz from 'jstz';
 export const DATE_FORMAT = 'yyyy-MM-dd';
 export const DATE_FORMAT_TZ = 'yyyy-MM-dd z';
 export const DATE_FORMAT_LONG = 'yyyy-MM-dd HH:mm:ss';
+export const DATE_FORMAT_LONG_NAME = 'dd MMM yyyy, HH:mm';
```
mattyg, line 8: The date format shown in the tooltip should include the day-of-week name, see mockup:
image

yarn.lock
```diff
@@ -8921,6 +8921,7 @@ __metadata:
     react-router-dom: ^5.2.1
     react-scripts: 5.0.0
     react-table: ^7.7.0
+    react-tooltip: ^4.2.21
```
mattyg, line 8924: This component has a nice look. If we want to use it though, we should update *all* tooltips in the application to use it as well. 

Currently I think the following components have tooltips (look for the attribute `title=`)
- InlineLabel
- EventLog
- ForwarderTooltip
- QuantifySlider

mattyg (changes requested):
Great! I requested some changes inline above. Also the tooltip didn't seem to appear when the Praise component is rendered within QuantifyTable.

mattyg (approved):
added changelog

PR #403, opened 2022-05-23
Bug/closing period UI reflection

nebs-dev:
Resolves #389 
Resolves #390 

- Fix UI reactivity when single period is updated - on close period and update period name
- in `useSinglePeriodQuery` `fetchPeriod` is not called on period change anymore, only if periodId is changed

PR #405, opened 2022-05-25
Fix: Prevent infinite fetching of period data

mattyg:
Fixes bug introduced in #403 where `useSinglePeriodQuery` would infinitely continue fetching a single period.


kristoferlund (approved):
👍👍👍

PR #406, opened 2022-05-25
Fix: Prevent Transitive Duplicates

### LINKED ISSUE #394: Prevent "transitive" duplicates
When marking a praise as duplicate, it must not be considered the "original" of another duplicate praise.

**Steps to reproduce**
On QuantifyTable, given the following praise in order: [ A, B, C ]
- Check praise B
- Check praise C
- click "mark duplicates"
- confirm
- Check praise A
- Check praise B
- click "mark duplicates"
- confirm

**Praise B will be a duplicate of praise A, and praise C will be a duplicate of Praise B**

mattyg:
Resolves #394 

**Overview**
- Throw error if user attempts to mark a praise as duplicate, when it is already marked as the original of another duplicate praise
- Minor refactoring of `quantify` controller


kristoferlund (approved):
Nice!

PR #407, opened 2022-05-25
Feat/display duplicate score

### LINKED ISSUE #378: Display calculated score of praise marked as duplicate
Currently when praise is marked as duplicate, the original praise slider is adjusted, while the duplicate sliders move to '0' and are disabled. 

This has caused some confusion for quantifiers. See this message from liviade on TEC discord:

> 1. marking duplicates using the new button isn't quite working. The value of the quantified praise goes to 0 when I select them both. 

We should make the UX of praise marked duplicate more clear by removing the slider entirely, and replacing it with a text box that indicates the calculated score it will receive.

mattyg:
Resolves #378

**Overview**
- merge changes from #406 
- display calculated duplicate score when praise marked duplicate
- hide button to mark a single praise as duplicate -- if praise is already a duplicate


kristoferlund (changes requested):
Not sure how to best resolve the merge conflict.

kristoferlund (dismissed):
Looks good!

kristoferlund (approved):

PR #409, opened 2022-05-25
Fix/update period settings

mattyg:
Fix issue where an error was thrown when attempting to update period settings created before migration `07_settings_add_group`.


kristoferlund (approved):
👍👍👍

PR #408, opened 2022-05-25
Fix/reactive quantifier pool size

mattyg:
Resolves #388

**Overview**
- Merge changes from #405 
- Refactor recoil architecture for period pool requirements -- store in atomFamily, fetch once via period Detail component
- Pass pool requirements into AssignDialog

PR #416, opened 2022-05-26
CRA 5.0.1 / Tailwind 3

kristoferlund:
## Frontend
- Upgrades Create React App to 5.0.1
- Upgrades Tailwind to v3

Might fix build errors in #408 and #413


mattyg (approved):
thanks for dealing with this!

PR #425, opened 2022-05-29
Prevent error notice when non-admin visits PeriodDetailPage

mattyg:
This prevents an unecessary failed api request for non-admin users when visiting PeriodDetailPage


kristoferlund (approved):
Nice! 👍

PR #437, opened 2022-06-02
bug/praise_reason_link

### LINKED ISSUE #427: Clicking a link within a Praise reason should *only* trigger opening the link
Currently, when clicking a link embedded in a Praise reason:
- the link is opened
- the Praise onClick is triggered

Instead, it should only trigger opening the link, not the Praise onClick

nebs-dev:
Resolves #427

- Created PraiseRow component that will handle row click on every page that we are showing praise


kristoferlund (approved):
Looks great! 🥳 Merging

PR #438, opened 2022-06-03
Fix/respect db username

kristoferlund:
When running the install script, manual database username changes in `.env` was being overwritten. Now the usernames are instead included as install questions.


mattyg (approved):
Good catch!

PR #439, opened 2022-06-03
Fix/no praise message

### LINKED ISSUE #273: Show message when there is no praise
When system first is installed, this is what you get:

image

Vyvy-vi:
resolves #273 
Screenshot 2022-06-03 at 11 16 38 PM


mattyg (changes requested):
There's a little quirk where the 'No praise has been dished yet' message appears temporarily while the praise data is loading. 

We should make it only appear when there is no praise *and* it's not in the middle of loading. I think this will likely require some refactoring of PraisePageLoader (maybe moving it into PraiseTable so we can access the `loading` state.

mattyg (approved):
Just noticed we have the same issue noted above on the MyPraiseTable. Maybe it makes sense to just merge this and create another ticket to address that issue in both components.

PR #440, opened 2022-06-03
Feat: Require discord role option

### LINKED ISSUE #419: Whoami command says user does not have praise powers, when they do
Reported by `@rex` on TEC discord:
> btw the whoami text says I don't have praise powers but I do, inconsequential bug but bug nonetheless

### LINKED ISSUE #434: Option - Require discord role to allow praise giving

kristoferlund:
Closes #434 
Closes #419 

- Add option to enable/disable the requirement to assign a certain Discord role to users before they can praise.
- Centralise the logic for the Discord role check to one function `assertPraiseGiver`.


packages/discord-bot/src/utils/assertPraiseGiver.ts
```diff
@@ -0,0 +1,53 @@
...
+  sendReply: boolean
+): Promise<boolean> => {
+  const praiseGiverRoleIDRequired = (await settingValue(
+    'PRAISE_GIVER_ROLE_ID_REQUIRED'
+  )) as boolean;
+  const praiseGiverRoleID = (await settingValue(
+    'PRAISE_GIVER_ROLE_ID'
+  )) as string;
+
+  if (praiseGiverRoleIDRequired) {
```
mattyg, line 17: Could we make this first conditional return a value, instead of wrapping all the others? Just a bit easier to read.

i.e.  
```
  if (!praiseGiverRoleIDRequired) {
   return true;
  }
  
  ...
  ```

mattyg (dismissed):
looks good! Just requested one minor change inline

packages/discord-bot/src/utils/assertPraiseGiver.ts
kristoferlund: 👍 done

Issue #435, opened 2022-06-01
Failed migration on update 2022-06-01 post-mortem 


mattyg:
## Failed Deployment Post-Mortem

- In #397 I created a migration to set `reasonRealized` for all existing praise -- by replacing discord's machine-readable formatting of users and channels with human-readable formatting
- The migration failed when it did not receive all successful responses from the discord api. (see image below)
- Because of this, the migration was replaced with another that set `reasonRealized` for all existing praise to `praise.reason`. 

### Log of failed migraiton

![img](https://cdn.discordapp.com/attachments/895054667921379331/981561527406592072/unknown.png)

### Resarch
- Migration failed on discord api request to fetch info on a channel `948663379487252530`
- The channel is a thread - `Omega > Consilience library` 
- Unable to run `/praise` command inside a thread currently (see image below)
![image](https://user-images.githubusercontent.com/159270/171515661-28694273-64f2-4e95-84cb-94de268cd952.png)

### Reflections

In the future we might want to change how we handle failed migrations:

- When a migration fails on deployment, revert to the previous deployment. Currently the process would be:
    1.  Revert to previous codebase
    2.  Run all migrations in reverse `yarn run migrator down` (for each new migration)

- Additionally, it might be easier if we *squashed* multiple migrations into one, for any release that has multiple migrations. That way, the process to revert to a previous deployment would be:
    1.  Revert to previous codebase
   (the database would already be unchanged due to the failed migration).

Vyvy-vi:
Some workarounds to fix the migrations -
1. Using the most openly accessible channel in the server when calling `Utils.cleanContent()`:
This ensures that the bot never has to query the channel in which praise was sent, resolving the issue we're having above. And. this also allows us to realise the content better by ensuring that all user ids get rendered irrespective of them having access to that channel.
OR
2. Write code to fetch the channels that can account for [different types of channels returned used in discord.js](https://github.com/discordjs/discord.js/issues/7321).
OR
3. Re-write `Utils.cleanContent` and use our own implementation that internally queries the data from the Guild object.

PR #442, opened 2022-06-03
Fix/praise reason realised refactor

Vyvy-vi:
Related to #435 

This PR resolves both the concerns -
1. Praise can now be used in threads again
2. Potentially fixes the 07_praise_reason_realized.ts migration.

Note - I didn't test this in a production environment.


packages/discord-bot/src/index.ts
```diff
@@ -39,7 +39,14 @@ discordClient.on('interactionCreate', async (interaction) => {
...
   try {
-    await command.execute(interaction);
+    if (
+      interaction.commandName === 'praise' ||
+      interaction.commandName === 'forward'
+    ) {
+      await command.execute(interaction, discordClient);
+    } else {
+      await command.execute(interaction);
+    }
```
mattyg, line 49: I think you can just drop this conditional entirely -- the functions will just ignore any additional parameters they're not expecting.

i.e.
```
  try {
    await command.execute(interaction, discordClient);
  } catch (error) {
  ...
  ```

File: packages/api/src/praise/utils/core.ts [hunk 2]
```diff
@@ -86,33 +86,47 @@ export const prepareDiscordClient = async (): Promise<Client> => {
...
-  if (!channel.isText()) throw Error('Channel must be a TextChannel');
-
-  const textRealized = Util.cleanContent(text, channel);
-
+  const discordGuild = await discordClient.guilds.fetch(
+    process.env.DISCORD_GUILD_ID as string
+  );
+  if (!discordGuild) throw Error('Failed to fetch guild from discord api');
+  await discordGuild.members.fetch();
+  await discordGuild.channels.fetch();
```
mattyg, line 108: Do we need to do all these fetches from the discord api on every call to `realizeDiscordClient`? 

Since the `DISCORD_GUILD_ID` is static, wouldn't the fetched members and channels be the same (at least for calls m? 

We'll need to be able to run `realizeDiscordClient()` many times in a batch (such as when we're importing a bunch of praise, or in a migration). 

Currently, when I tried to create a new migration that calls this function for every praise.reason, it runs for about 2 minutes and then fails due a time out from discord api. I'd imaging I'm running to a rate limit or something.

mattyg (changes requested):
awesome!

I requested some changes inline.

What was the underlying issue causing it to fail inside threads?

Vyvy-vi:
The `/praise` command failed in threads as `Util.cleanContent()` method wasn't able to fetch the channel being passed by the interaction, which was of `ThreadChannel` type.

packages/api/src/praise/utils/core.ts
mattyg: Yeah I think that makes sense -- try to use the value in the cache, and it fails, query the api.

Vyvy-vi: there's a small issue .replace doesn't support async functions.
Should we implement our own implementation that supports async-await, or should we use the `.then().catch()` syntax?

__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Summary:** :hammer_and_wrench: This code needs a fix

mattyg: > there's a small issue .replace doesn't support async functions. Should we implement our own implementation that supports async-await, or should we use the `.then().catch()` syntax?

Could we collect the list of user names, then confirm we have them cached in the discordClient or fetch them, then run the `.replace` since we would know we had all the data already?

Vyvy-vi: this could lead to outdated data being used when the bot's `/praise` or `/forward` commands are used.

__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Tags:** Not reusable

PR #453, opened 2022-06-06
WIP: Dark mode feature

### LINKED ISSUE #420: dark mode for praise dashboard
for those of us who stay up late at night quantifying praise, it would be cool to have a dark mode available. Right now it is very bright and white! 

This was given as some feedback from quantifiers during the previous review call.

pkretzschmar:
Fix #420

### DIFF
--- a/packages/frontend/src/index.tsx
+++ b/packages/frontend/src/index.tsx
@@ -4,13 +4,14 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { Toaster } from 'react-hot-toast';
 import { BrowserRouter as Router } from 'react-router-dom';
-import { RecoilRoot } from 'recoil';
+import { RecoilRoot, useRecoilValue } from 'recoil';
 import RecoilNexus from 'recoil-nexus';
 import { useErrorBoundary } from 'use-error-boundary';
 import EthConnection from './components/EthConnection';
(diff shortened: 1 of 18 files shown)


kristoferlund:
Works great! Finally! This is a highly requested feature, the quantifiers will be very happy once this is done!

Login screen (and most likely activation screen):

image

Here, some color adjustments (don't mind the layout): 

image

For the menus, use just one shade lighter for hover and active states, in this case: `bg-slate-800`.

In general, use one shade lighter for hover and active, throughout the dashboard.

image

Form fields look better if they are slightly lighter than background - `slate-300` in this case:

image

If main menu and submenus have different backgrounds then the hover bg needs to be different as well. One shade lighter than background.

image

Quantify header and sliders needs dark mode: 

image

For the switch, do a sun / moon icon switch similar to https://givepraise.xyz. This will later be moved to top navigation.

image

kristoferlund:
If possible we should make the Rainbow wallet connect dialog dark mode as well. Perhaps it has some built in support for this? 

image

kristoferlund (changes requested):
Added some styling comments. Overall, looks great!!

kristoferlund (approved):
Overall, looks great! I'm merging this now. We will need to revisit some things. I'll make separate issues for those.

PR #455, opened 2022-06-07
User Event Logs - filter, search and sorting

### LINKED ISSUE #376: Custom filtering, sorting, and searching of UserEventLogs
In #309 we implemented basic user event logging.

Currently the UX of user event logs is simply a paginated list, sorted by `UserEventLog.createdAt` descending.

For improved utility and UX, we should allow custom filtering by the following:
- `UserEventLog.type` (one or multiple)
- sorting by `createdAt` ascending or descending
- full-text search of all fields

nebs-dev:
Resolves #376 
- Added sorting, filtering and search UI
- Refactor Event Logs query and EventLogsTypes recoil methods
- Added EventLogsTypes transformer and controller method

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Filter, search and sort the event log #376 #455
 - Dark mode! #420 #453
 - Support multiple wallets: WalletConnect, Trust, Rainbow etc #424
 - Export the user list as csv #402 #450


packages/frontend/src/model/eventlogs.ts
```diff
@@ -1,49 +1,87 @@
...
+  get: ({ get }): AxiosResponse<unknown> => {
+    const query = get(eventLogsQueryParameters);
+
+    const qs = Object.keys(query)
+      .map((key) => `${key}=${query[key]}`)
+      .join('&');
+
+    const response = get(
+      ApiAuthGet({
+        url: `/eventlogs/all${qs ? `?${qs}` : ''}`,
```
mattyg, line 44: It might be a bit cleaner to use the `apiAuthClient` instead of the  `ApiAuthGet` recoil selector --  then you can pass the query parameters as an object and you don't have to do this string formatting stuff.

packages/frontend/src/pages/EventLogs/components/EventLogsActions.tsx
```diff
@@ -0,0 +1,198 @@
...
+  const [searchValue, setSearchValue] = useState<string>('');
+
+  const { types } = useAllEventTypeLogs();
+  const filterOptions = types.docs;
+
+  const [queryParameters, setQueryParameters] = useRecoilState(
+    eventLogsQueryParameters
+  );
+
+  const returnValues = useMemo(
```
mattyg, line 44: getting a lint error

mattyg (changes requested):
Awesome! 

I requested some minor changes inline. 

The only other change I'd request is that we style the search, filter, and sort input components to reflect the styling used in the new Users page. Ideally by making generic input components we can keep using throughout the application.

nebs-dev:
> throughout

ok.. i will have to check new user page and use those input styles

kristoferlund:
How about this:

image

- Skip the filter icon, it is vague. Go with the standard dropdown icon instead.
- The headless ui dropdown provides the checkboxes the default dropdown doesn't include. Let's make this component our default choice.
- Margins need to be adjusted, use margin `4` between components. 
- These inputs should be same height as a std text input (on my screen 42 px) - adjust margins pls. 

We'll look into styling borders etc soon - perhaps making them slightly more soft.

kristoferlund:
Great, almost there! I'll post a few comments before Monday.

packages/frontend/src/pages/EventLogs/components/EventLogsActions.tsx
mattyg: > hm, i am not sure what to do here - if i add dependencies for useMemo hook (queryParameters & setQueryParameters) i will create infinite loop because we are updating queryParameters here and they will change every time. Also i am not really sure why component is rerendering every time I update queryParameters @mattyg

Sorry I think I didn't look closely enough at this the first time. I think the confusion comes from using this pattern where EventLogsList reads+writes to recoil state, and EventLogsActions also reads+writes to recoil state, and the parameters to the api request are stored in recoil state.

I think it would be clearer if instead:
- EventLogsList was the "smart" component, that reads+writes to recoil state, and stores the query parameters *internally* in it's component state
- while EventLogsActions was a "dumb" component, that purely takes in user input and emits it to it's parent component.  Or alternatively, could be simply merged into it's parent component.

Does that make sense? We could pair program it too if you want.

kristoferlund (approved):
Great! I made some minor styling changes. Merging now.

kristoferlund (approved):

PR #463, opened 2022-06-12
fix: link overflow truncate

### LINKED ISSUE #457: Prevent text from overflowing Praise component content area
See screenshot below:
image

### LINKED ISSUE #465: Praise reasons containing URLs that including certain symbols are not rendered properly
Currently, if a praise reason includes a URL that includes a markdown symbol it is not rendered properly -- any markdown symbols that are included in the url are tread as such, which breaks the url -> link rendering.

See screenshots below:
image
image

Vyvy-vi:
Fixes #457
Fixes #465
this truncates long URLs to 50-character long hyperlinks.
and takes care of symbols inside hyperlinks.

Screenshot 2022-06-12 at 7 02 36 AM
Screenshot 2022-07-04 at 1 54 37 PM
Screenshot 2022-07-04 at 1 52 10 PM


mattyg (dismissed):
awesome thanks vy!

couple issues:
- If the viewport is too small, a 50 char URL will still overflow the content area. We need to add a CSS rules to break words if necessary. See the screenshot below. I think this is the CSS rule that resolves it: https://tailwindcss.com/docs/word-break#break-words

image

- lets consolidate this logic to render url strings to html links outside of the Praise component. We should be able to keep all the relevant logic in either `components/MarkdownText.tsx` or `utils/parser.ts` (we might want to consolidate them too).

- We have some redundant duplication of logic: in `components/MarkdownText.tsx` we use the `linkifyHtml()` function to convet url strings to html links. In this PR we're converting them twice (again in in `parser.tsx`), and then we're truncating any urls in `Praise.tsx`. Can we consolidate this into one implementation?

- If we want to keep using `linkifyHtml()`, it has a built-in option for handling truncation. See the option "format" on https://linkify.js.org/docs/options.html  - might be simpler code but I don't have a strong preference.

- While looking at this I noticed another bug that it probably makes sense to address in the same PR, since they related to the same code. The bug is that if a url contains any markdown symbols, the html link rendering is broken. See #465

nebs-dev:
Nice job.. for some reason `discord-markdown` didn't work for me and that is why i created custom parser, this is even better.

nebs-dev (approved):

PR #464, opened 2022-06-12
Feat/disable self praise

### LINKED ISSUE #431: Option to disable self praise

Vyvy-vi:
Resolves #431
This PR adds the following  -
- Minor refactors
- Prevent duplicate praise when user mentioned twice in receivers field of `/praise` or `/forward`
- Option to disable self-praise in the `/praise` and `forward` commands


packages/discord-bot/src/utils/praiseEmbeds.ts
```diff
@@ -208,3 +208,11 @@ export const notActivatedDM = async (msgUrl: string): Promise<MessageEmbed> => {
     `**[YOU HAVE BEEN PRAISED](${msgUrl})\nPRAISE ACCOUNT NOT ACTIVATED. USE \`/activate\` TO ACTIVATE YOUR ACCOUNT. (message not set)`
   );
 };
+
+export const selfPraiseWarning = async (): Promise<string> => {
+  const msg = (await settingValue('SELF_PRAISE_WARNING')) as string;
+  if (msg && typeof msg === 'string') {
+    return msg;
```
mattyg, line 215: we shouldn't need to check `typeof msg` since it gets cast to a string immediately before.

packages/discord-bot/src/handlers/praise.ts
```diff
@@ -63,44 +57,44 @@ export const praiseHandler: CommandHandler = async (
...
     return;
   }
 
   const praised: string[] = [];
-  const receiverIds = receiverData.validReceiverIds.map((id: string) =>
-    id.replace(/\D/g, '')
-  );
+  const receiverIds = [
+    ...new Set(
+      receiverData.validReceiverIds.map((id: string) => id.replace(/\D/g, ''))
```
mattyg, line 70: :+1: good call to remove duplicates!

mattyg (changes requested):
It looks like if I only praise myself (and no other users), it responds with two errors: both 'Receivers Not Mentioned' and 'Self Praise Not Allowed'. Can we make it only respond with the "Self Praise Not Allowed.

I think it's okay if it responds with "Self Praise Not Allowed" and fails to create the praise if my user account is *any* of the receiver.

Vyvy-vi:
> It looks like if I only praise myself (and no other users), it responds with two errors: both 'Receivers Not Mentioned' and 'Self Praise Not Allowed'. Maybe it could only respond with the "Self Praise Not Allowed.
> 
> I think it's okay if it responds with "Self Praise Not Allowed" and fails to create any praise if my user account is within the receivers.

We could set that up :+1:

__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Summary:** :hammer_and_wrench: This code needs a fix

packages/discord-bot/src/utils/praiseEmbeds.ts
```diff
@@ -211,7 +211,7 @@ export const notActivatedDM = async (msgUrl: string): Promise<MessageEmbed> => {
 
 export const selfPraiseWarning = async (): Promise<string> => {
   const msg = (await settingValue('SELF_PRAISE_WARNING')) as string;
-  if (msg && typeof msg === 'string') {
+  if (msg) {
```
mattyg, line 214: actually I think we don't need any of these `if(msg)` statements. settingValue will never return undefined -- instead it throws an error if the setting isn't found.

mattyg (approved):
Noted one more possible change. Otherwise looks good!

packages/discord-bot/src/utils/praiseEmbeds.ts
Vyvy-vi: So, should we add a try-catch block here?

__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Summary:** :hammer_and_wrench: This code needs a fix&nbsp; | &nbsp;**Tags:** Inefficient

mattyg: > So, should we add a try-catch block here?

hmm, it depends how we want to handle the "default" value for `selfPraiseWarning`. If we want to keep it statically defined in the codebase as it is now, then yes wrapping it in a try-catch makes sense (and in the catch we would return the default value). 

Though I think it might be simpler to just remove the `selfPraiseWarning` wrapper entirely. If the setting is not defined it throws an error, which bubbles up to the handler function, which catches it and displays some user-facing message like "failed to forward praise".

does that make sense? It could also go in a separate PR as a more general refactor of the pattern in all the places it's used? Maybe that makes more sense.

PR #479, opened 2022-06-19
WIP Test: additional api tests

mattyg:


kristoferlund:
> * modify application seeder so all periods and praise are from the _past_, instead of the future

Yes! This has been annoying me slightly for quite some time.

PR #498, opened 2022-06-22
Fix: Prevent error when not all quantifiers assigned

### LINKED ISSUE #492: Unable to assign quantifiers using PRAISE_QUANTIFIERS_ASSIGN_EVENLY
`ERROR: InternalServerError: Not all quantifiers were assigned praise, missing 1, despite PRAISE_QUANTIFIERS_ASSIGN_EVENLY`

mattyg:
Resolves #492 

**Overview**
- Remove check that was throwing an error when not all quantifiers were assigned praise and `PRAISE_QUANTIFIERS_ASSIGN_ALL` was enabled
- Change key, label & description of setting `PRAISE_QUANTIFIERS_ASSIGN_ALL` to clarify that it is not always possible to distribute praise assignments among *all* quantifiers (for example, if the number of receivers in a period is less than the number of quantifiers).
- Fix setting and periodsetting tests -- generate fake settings to test with rather than expecting migration's "real" settings to exist


kristoferlund (dismissed):
Looks good!

PR #499, opened 2022-06-23
keyword search for quantifiers & Quantify multiple praise

### LINKED ISSUE #422: keyword search for quantifiers
I LOVE this new header for marking duplicates! 

gathering feedback from quantifiers it would be awesome to implement a keyword search to make marking duplicates and scoring easier. 

having a text field right in the duplicate/dismiss header that users could input a keyword and it would highlight all matching text on the page. marking duplicates or dismissing or doing any interactions on the page would not reset or remove the keywords input.

image

nebs-dev:
Resolves #422

- Added search by praise reason
- Added option so quantify multiple praise at once

### DIFF
--- a/packages/api/src/praise/controllers.ts
+++ b/packages/api/src/praise/controllers.ts
@@ -246,6 +246,5 @@ export const quantifyMultiple = async (
   );
 
   const response = await praiseDocumentListTransformer(affectedPraises);
-  console.log('RESPONSE:', response);
   res.status(200).json(response);
 };
--- a/packages/api/src/praise/controllers.ts
+++ b/packages/api/src/praise/controllers.ts
@@ -31,6 +31,7 @@ import {
   PraiseDocument,
   PraiseDto,
   QuantificationCreateUpdateInput,
+  QuantifyMultiplePraiseInput,
 } from './types';
 import { praiseWithScore, getPraisePeriod } from './utils/core';
 
@@ -199,5 +200,52 @@ export const quantify = async (
   );
 
   const response = await praiseDocumentListTransformer(affectedPraises);
+  res.status(200).json(response);
+};
+
+export const quantifyMultiple = async (
+  req: TypedRequestBody<QuantifyMultiplePraiseInput>,
+  res: TypedResponse<PraiseDto[]>
+): Promise<void> => {
+  const { score, praiseIds } = req.body;
+
+  const affectedPraises = await Promise.all(
+    praiseIds.map(async (id) => {
+      const praise = await PraiseModel.findById(id).populate(
+        'giver receiver forwarder'
+      );
+
+      if (!praise) throw new NotFoundError('Praise');
+
+      const period = await getPraisePeriod(praise);
+
+      if (!period)
+        throw new BadRequestError('Praise does not have an associated period');
+
+      if (!res.locals.currentUser?._id) {
+        throw new InternalServerError('Current user not found.');
+      }
+
+      const quantification = praise.quantifications.find((q) =>
+        q.quantifier.equals(res.locals.currentUser._id)
+      );
+
+      if (!quantification)
+        throw new BadRequestError(
+          'User not assigned as quantifier for praise.'
+        );
+
+      quantification.score = score;
+      quantification.dismissed = false;
+      quantification.duplicatePraise = undefined;
+
+      await praise.save();
+
+      return praise;
+    })
+  );
+
+  const response = await praiseDocumentListTransformer(affectedPraises);
+  console.log('RESPONSE:', response);
   res.status(200).json(response);
 };
--- a/packages/frontend/src/components/form/SearchInput.tsx
+++ b/packages/frontend/src/components/form/SearchInput.tsx
@@ -3,15 +3,17 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 interface SearchInputProps {
   handleChange: (element) => void;
-  value: string;
+  value?: string;
+  placeholder?: string;
 }
 
 const SearchInput = ({
   handleChange,
   value,
+  placeholder = 'Search',
 }: SearchInputProps): JSX.Element => {
   return (
-    <div className="relative flex items-center border border-warm-gray-400 h-[42px]">
+    <div className="relative flex items-center border border-warm-gray-400">
       <div className="absolute inset-y-0 left-0 flex items-center pl-3">
         <span className="text-warm-gray-800 dark:text-white">
           <FontAwesomeIcon
@@ -25,7 +27,7 @@ const SearchInput = ({
         className="block pl-8 bg-transparent border-none outline-none focus:ring-0"
         name="search"
         type="text"
-        placeholder="Search"
+        placeholder={placeholder}
         value={value}
         onChange={(e: React.ChangeEvent<HTMLInputElement>): void =>
           handleChange(e.target.value)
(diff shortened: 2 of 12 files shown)


packages/frontend/src/pages/QuantifyPeriodReceiver/components/QuantifyTable.tsx
```diff
@@ -110,20 +163,55 @@ const QuantifyTable = ({ periodId, receiverId }: Props): JSX.Element | null => {
...
+                  checked={selectAllChecked}
+                />
+                <span className="ml-4">Select All</span>
+              </td>
+            </tr>
+            <tr>
+              <td colSpan={5}>
+                <div className="mb-5 border-2 border-t border-warm-gray-400" />
+              </td>
+            </tr>
```
mattyg, line 213: What do  you think about moving the 'Select All' checkbox into the "toolbar" area above? That way it will be "sticky" when user scrolls down the page. Also then we could remove this additional border line. 

Something like this:
Screenshot_2022-06-26_12-43-46

packages/frontend/src/pages/QuantifyPeriodReceiver/components/QuantifyMultipleButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+    <button
+      disabled={disabled}
+      className={
+        disabled
+          ? 'praise-button-disabled space-x-2'
+          : 'praise-button space-x-2'
+      }
+      onClick={onClick}
+    >
+      <FontAwesomeIcon icon={faPrayingHands} size="1x" />
```
mattyg, line 23: What do you think about using the "unbalanced scale" icon instead of praying hands? https://fontawesome.com/search?q=scale%20unbalanced&s=solid%2Cbrands%2Cregular

packages/frontend/src/pages/QuantifyPeriodReceiver/components/QuantifyMultipleDialog.tsx
```diff
@@ -0,0 +1,63 @@
...
+          <button className="praise-button-round" onClick={onClose}>
+            <FontAwesomeIcon icon={faTimes} size="1x" />
+          </button>
+        </div>
+        <div className="px-20 space-y-6">
+          <div className="flex justify-center">
+            <FontAwesomeIcon icon={faCalculator} size="2x" />
+          </div>
+          <h2 className="text-center">
+            Quantify selected ({selectedPraises.length}) praises
```
mattyg, line 39: Should we list the id's of the praise items that will be affected, like in the DismissDialog? (I don't feel strongly I just think they should be consistent).
Screenshot_2022-06-26_12-52-23

packages/frontend/src/pages/EventLogs/EventLogsPage.tsx
```diff
@@ -109,7 +109,7 @@ const EventLogsPage = (): JSX.Element => {
           </div>
 
           {/* Search */}
-          <div className="w-5/12 mt-5 mb-5 mr-4">
+          <div className="w-5/12 mt-5 mb-5 mr-4 h-[42px]">
```
mattyg, line 112: Can we use tailwind's built-in sizing classes instead of fixed pixel sizes?

packages/frontend/src/pages/EventLogs/EventLogsPage.tsx
```diff
@@ -120,7 +120,7 @@ const EventLogsPage = (): JSX.Element => {
           </div>
 
           {/* Sort */}
-          <div className="w-2/12 mt-5 mb-5 ml-auto mr-5">
+          <div className="w-2/12 mt-5 mb-5 ml-auto mr-5 h-[42px]">
```
mattyg, line 123: same as above

packages/frontend/src/components/form/SearchInput.tsx
```diff
@@ -25,7 +27,7 @@ const SearchInput = ({
         className="block pl-8 bg-transparent border-none outline-none focus:ring-0"
         name="search"
         type="text"
-        placeholder="Search"
+        placeholder={placeholder}
```
mattyg, line 30: :+1:

packages/api/src/praise/controllers.ts
```diff
@@ -201,3 +202,49 @@ export const quantify = async (
   const response = await praiseDocumentListTransformer(affectedPraises);
   res.status(200).json(response);
 };
+
+export const quantifyMultiple = async (
```
mattyg, line 206: This function will need to have same logic used in the `quantify` function to determine *all* the praises with a `scoreRealized` affected by the changes. It probably makes sense to pull that logic out into a utility function and use it in both controllers.

The reason for that logic is so the "Duplicate Score" display is updated when the original praise's score is updated. See the screencast below:

https://user-images.githubusercontent.com/159270/175825820-fac4dde2-e2bc-4d8f-8bcf-2441e9875781.mp4

mattyg (changes requested):
Awesome this looks great!

I requested some changes inline. 

Also a few additional I'd like to request:
- [ ] After dismissal, quantifying multiple, or marking duplicates, the "Select All" checkbox itself is unchecked
- Write tests for the new quantifyMultiple controller (you can do this in another PR -- you'll mostly copy the praise controller tests from pr #479)

packages/frontend/src/pages/QuantifyPeriodReceiver/components/QuantifyTable.tsx
```diff
@@ -31,21 +38,40 @@ const QuantifyTable = ({ periodId, receiverId }: Props): JSX.Element | null => {
...
     'PRAISE_QUANTIFY_ALLOWED_VALUES'
   ) as number[];
 
+  const applyFilter = React.useCallback(
+    (data: PraiseDto[] | undefined): PraiseDto[] => {
+      if (!data) return [];
+      if (!searchValue) return data;
+
+      const filteredData = data.filter((praise: PraiseDto) => {
+        return praise.reasonRealized.includes(searchValue);
```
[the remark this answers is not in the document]
nebs-dev: and which username should we use? receiver's or giver's?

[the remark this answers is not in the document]
mattyg: how about both?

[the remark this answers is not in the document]
kristoferlund: Yess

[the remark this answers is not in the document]
kristoferlund: Yesss

packages/frontend/src/pages/EventLogs/EventLogsPage.tsx
kristoferlund: We just want to ensure all the inputs are the same height. If that can be achieved without specifying a fixed height, all the better.

kristoferlund (changes requested):
I believe it is time to remove the text from the buttons. 


- Move "select all" checkbox to inline with buttons
- Hide and show only relevant buttons. No items checked = no buttons visible.
- Use the `praise-button-round` style
- Tooltip below buttons describe function
- (see mail.google.com for reference 😄)
- Please note that image below is just wireframe, use exisiting styles and colours.

image

kristoferlund (approved):
I added the `IconRoundButton` because I wanted another look than the regular button. And had to make a custom tooltip because the MUI one don't work over well buttons that are disabled.

We can talk the changes through tomorrow if you like?

Merging now.

PR #503, opened 2022-06-26
Feat: replace active quantifier

### LINKED ISSUE #432: Allow admins to replace quantifiers

mattyg:
Resolves #432 

**Overview**
- refactor button components MarkDismissedButton, MarkDuplicateButton to use a generic IconButton
- refactor UserPopover to no longer use headlessui's Popover (as it only supports click-based popover and wasn't being utilized)
- Resolve #432
- Remove cli command `replace-quantifier`

Note that we are currently including *all* quantifiers in the list of possible replacements, not just those currently unassigned.


mattyg:
Currently, only unassigned quantifiers appear in the list of possible replacements. Should we instead include *all* quantifiers in the list of possible replacements?

@kristoferlund

mattyg:
> Currently, only unassigned quantifiers appear in the list of possible replacements. Should we instead include _all_ quantifiers in the list of possible replacements?
> 
> @kristoferlund

We decided to include *all* quantifiers for now.

kristoferlund (changes requested):
Looks great! 

One thing needs fixing though. Allowing the choice of any quantifier can cause a situation like this:

image

Perhaps we need to check if the new quantifier already have been assigned any of the receivers of the quantifier being replaced?

mattyg:
> Perhaps we need to check if the new quantifier already have been assigned any of the receivers of the quantifier being replaced?

good catch! Fixed.

Also made a few other changes:
- Exclude the original quantifier from the list of possible replacements
- Clear the selected user from the SelectUserRadioGroup after clicking "Replace" button
- Update success message to clarify that the scores are reset
- added tests for replaceQuantifier controller

kristoferlund (approved):
Great fixes! Merging.

PR #519, opened 2022-07-20
move user table params to the recoil state

### LINKED ISSUE #494: The User page should remember state when navigating to the User detail page and back

nebs-dev:
Resolves #494

### DIFF
--- a/packages/frontend/src/pages/Users/components/UsersTable.tsx
+++ b/packages/frontend/src/pages/Users/components/UsersTable.tsx
@@ -2,10 +2,7 @@ import React from 'react';
 import { useRecoilValue, useRecoilState } from 'recoil';
 import { UserDto, UserRole } from 'api/dist/user/types';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import {
-  faCircleXmark,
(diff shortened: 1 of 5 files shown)


packages/frontend/src/pages/Users/components/UsersTable.tsx
```diff
@@ -1,22 +1,24 @@
...
 import { SelectInput } from '@/components/form/SelectInput';
 import { UsersTableRow } from './UsersTableRow';
 import { UsersTablePagination } from './UsersTablePagination';
 
-interface roleOptionsProps {
-  value: string;
-  label: string;
-}
-
 const roleOptions = [
```
kristoferlund, line 22: You are maintaining two roleOptions lists. One here and one in `model/users.ts`. One should be enough. Consider removing this list and exporting the one found in `model`.

kristoferlund (changes requested):
Works great!

I had one comment regarding duplication of `roleOptions`.

And, add the x icon to the `SearchInput` component, inside the text input frame. similar to how the magnifying glass is done but opposite side.

image

kristoferlund (approved):
Looks good!

PR #520, opened 2022-07-21
Remove statically-defined users from application seeder #500

### LINKED ISSUE #500: Remove statically-defined users from application seeder
In #479 the seeder functions were refactored significantly. 

Continuing on that work, we should remove the `PREDEFINED_USERS` from `seeder/app.ts`, instead relying on the `seedAdminUsers()` function which generates admin users from `process.env.ADMINS`.

This ensures that our pre-defined set of admins (from the TEC and praise dev team) are not automatically made admins in other communities production deployments.

nebs-dev:
Resolves #500


kristoferlund (approved):
Great!

Issue #459, opened 2022-06-09
Bot message to first time praisers


kristoferlund:
“We noticed this is the first time you praise, here is some good info about what makes good Praise”!

PR #521, opened 2022-07-23
fix: ephemeral warnings

### LINKED ISSUE #496: Some praise bot messages that currently are public should most likely be ephemeral
image

Vyvy-vi:
Partially fixes #496

This PR turns the warnings sent after the initial praise message into one ephemeral reply.
Screenshot 2022-07-23 at 8 27 58 PM
Screenshot 2022-07-23 at 8 27 33 PM

---

While this solves the issue for the warnings sent after the first praise bot response(success / failure), the initial error messages can't be made ephemeral in a neat way as we want variable behaviour from that message:
1. We want to defer it so we have 15 minutes to run the praise logic. At this stage, we need to specify if that message is ephemeral. 
2. If the praise process is successful we want to edit the temporary deferred reply with new message content, which is publicly visible.
3. If the praise process is unsuccessful we want to edit the temporary deferred reply with new message content displaying errors, which is private(ephemeral).

The can't change the state of an already deferred message. A potential solution can be - Not updating the deferred reply with the error and send a response like:
```
🤞 <@praiseGiver>, we were unable to praise <@receivers[]> for <reason>
```
and then send an ephemeral followup with the error details that is private. 

Would the above be a desirable solution, or should we refactor the way the bot sends responses(~~might turn into a mess~~)?


Vyvy-vi:
this has to be merged after #459 is merged in. 

__
[sema-logo](https://semasoftware.com/gh) &nbsp;**Tags:** Not reusable

kristoferlund (approved):
Looks good! Merging.

PR #533, opened 2022-07-29
try to assign 3 times

### LINKED ISSUE #497: Several attempts to assign praise to quantifiers
Currently the architecture of quantifier assignment is roughly:
- Randomly group praise into "buckets"
- Randomly assign a quantifier to each "bucket"
- Verify that all bucket assignments are valid (i.e. no quantifier is assigned to their own praise, and all praise is accounted for)

Because of this randomness and algorithm, it is possible for an assignment to fail once, and then succeed another time, without changing the praise or adding quantifiers.

There is likely a better algorithm that could eliminate this issue. However, it may be a simple short-term improvement to retry quantifier assignment several times, if preceding attempts fail.

nebs-dev:
Resolves #497


kristoferlund (approved):
Nice! Merging.

PR #537, opened 2022-08-06
Replace praise-* class names with React components

### LINKED ISSUE #482: Replace `praise-*` class names with React components
@pkretzschmar, this is part of a larger refactor of the frontend. Replacing css classes with React components is in many cases a good idea, for example:

Instead of css/Tailwind classes:
```
praise-button
praise-button-outline
```
Create React components:
```
<Button/>
<Button outline/>
```

`praise-box` and `praise-page` can also be refactored. 

See Tailwind tips here: https://tailwindcss.com/docs/reusing-styles#extracting-components-and-partials

Another refactor is: encapsulating common patterns in more generic React components. `Dialog` from '@headlessui/react' for example is used in some places but used slightly different each time. 

What other components could be made generic?

Pinging @nebs-dev @mattyg and @Vyvy-vi as well.

nebs-dev:
Resolves #482 

- created new general components - PraiseButton, PraisePage, PraiseBox
- moved UI components to components/ui folder


packages/frontend/src/components/ui/PraiseButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+  id,
+  type,
+  variant = '',
+  classes,
+  children,
+  onClick,
+}: PraiseButtonProps): JSX.Element => {
+  const isDisabled = variant === 'disabled';
+  const variantClass =
+    variant === '' || !variant ? 'praise-button' : `praise-button-${variant}`;
```
kristoferlund, line 29: We should avoid creating classnames dynamically. See Tailwind docs for reference:

https://tailwindcss.com/docs/content-configuration

image

kristoferlund (changes requested):
- Using "Praise" in the component names is superfluous, and it is also not being applied consistently. Rename `PraiseButton` to `Button` and so on.
- Remove unused css classes such as `praise-button-round`
- Make sure the right components are used consistently. `UserStatistics.tsx` still uses `<div className="praise-button">` for instance.

packages/frontend/src/components/ui/PraiseButton.tsx
kristoferlund: Also, avoid using the global classnames using `@apply`, especially when they are only used in one component. Instead, define local variables for the variants:

```
const outline = {
  px-4 py-2 …
}
```

kristoferlund (changes requested):

kristoferlund (approved):
Great! Merging…

PR #541, opened 2022-08-10
save app version change message key to local storage

### LINKED ISSUE #528: Allow user to close sticky version message
- Close icon `x` aligned way to the right
- State should be persisted so notification does not reappear on reload
- But .. if github version number increases again then notification should reappear

nebs-dev:
Resolves #528


kristoferlund (approved):
Merging!

PR #549, opened 2022-08-15
Custom export format

### LINKED ISSUE #532: Custom export formats
Custom export formats will allow communities to create token distribution files etc without having to rely on separate tools such as RAD.

To begin with we will make a simple setup where we allow for **one** custom export format. 

A custom export takes the raw period receiver data and transforms it into another format. The custom format transformer (map) should also accept parameters (context) such as:
- The token distribution budget
- Token name to distribute
- ... any context parameter

I believe we might be able to use this library to perform the transform: https://github.com/bozzltron/node-json-transform

## Specs
- Raw data is the summarized period receiver data. No individual praise items included but the summarised score for the period in question.
- Three global settings: 
  - `CUSTOM_EXPORT_MAP`: url pointing to a JSON object containing the export map/transformer. We will want to setup a separate repo for the transformers to allow for the community to contribute new transformers.
  - `CUSTOM_EXPORT_CONTEXT`: JSON object with the default inputs used by the transform
  - `CUSTOM_EXPORT_CSV_FORMAT`: Boolean
    - true: Export in csv format
    - false: Export in JSON format 
- Period settings: map, context and format:
  - Period settings inherit and override global defaults
  - These settings should not be locked when quantification has begun
- Button on the period summary page: `Custom export`

## First transformer

Let's start with creating a transform to the csv format needed by Aragon Finance.

Output CSV:

```
[ADDRESS], [AMOUNT], [TOKEN NAME]
[ADDRESS], [AMOUNT], [TOKEN NAME]
[ADDRESS], [AMOUNT], [TOKEN NAME]
```

Context:
```
{
  budget: 1000,
  token: 'TEC'
}
```

- `ADDRESS`: Praise receiver address
- `AMOUNT`: ` praiseScore / context.totalPraiseScore * context.budget`
- `TOKEN`: `context.token`

`totalPraiseScore` is added to the context by the transformer at runtime along with other parameters that might be needed by the transform: `receiverCount`, … what else?

nebs-dev:
Resolves #532 

Added:
- new export options for the period token distribution
- new settings section for custom export
- api endpoint `/exportSummary` that i using export data transformer from any external url
- refactor of settings form


packages/frontend/src/pages/Settings/SettingsPage.tsx
```diff
@@ -53,6 +56,10 @@ const SettingsPage = (): JSX.Element | null => {
               <NavItem to={`${url}/application`} description="Application" />
               <NavItem to={`${url}/period`} description="Period Defaults" />
               <NavItem to={`${url}/discord`} description="Discord Bot" />
+              <NavItem
```
kristoferlund, line 59: The cusom export settings page doesn't render properly here. 

"Rendered more hooks than during the previous render."

packages/frontend/package.json
```diff
@@ -39,9 +39,11 @@
     "env-cmd": "^10.1.0",
     "ethers": "^5.6.8",
     "final-form": "^4.20.4",
+    "js-base64": "^3.7.2",
```
kristoferlund, line 42: Currently unused.

packages/api/src/period/types.ts
```diff
@@ -124,3 +125,9 @@ export interface ReplaceQuantifierRequestBody {
   currentQuantifierId: string;
   newQuantifierId: string;
 }
+
+export interface SummarizedPeriodData {
```
kristoferlund, line 129: This represents one interface for one custom export only. There will be many export types defined by .. whoever. There will be no export intefaces since everything is based just on the transformer json format.

packages/api/src/settings/validators.ts
```diff
@@ -89,6 +90,11 @@ export function isSettingValueAllowedBySettingType(
     return valid;
   }
 
+  if (this.type === 'Object') {
```
kristoferlund, line 93: It is not `object` but `JSON`, right?

packages/frontend/package.json
```diff
@@ -58,6 +60,7 @@
     "recoil": "^0.7.4",
     "recoil-nexus": "^0.4.0",
     "recoil-persist": "^4.2.0",
+    "safe-eval": "^0.4.1",
```
kristoferlund, line 63: Do `safe-eval` need to be imported in frontend?

kristoferlund (changes requested):

packages/frontend/src/model/periods.ts
```diff
@@ -488,6 +490,41 @@ export const useExportPraise = (): useExportPraiseReturn => {
...
+
+  const exportSummaryPraise = async (
+    period: PeriodDetailsDto,
+    exportContext: string,
+    supportPercentage: boolean
+  ): Promise<Blob | undefined> => {
+    if (!period) return undefined;
+
+    const response = await apiAuthClient.get(
+      `/admin/periods/${period._id}/exportSummary?context=${exportContext}&supportPercentage=${supportPercentage}`,
```
kristoferlund, line 515: `/exportSummary` returns summarised praise data for the period, basically the input to the custom export
New endpoint: `/customExport` does what `/exportSummary` currently does.

kristoferlund:
@nebs-dev remove checkbox in the export dialog and instead just print message. 

If CS support percentage > 0:

Thank your for supporting the continued development of Praise, **2%** of the distribution will be donated to the Praise dev team.

If CS support percentage = 0:

Consider supporting the development of Praise by donating a small amount of the distribution to the Praise dev team. [Settings: Custom token export](/linktosettings)

kristoferlund (approved):
Great work with this issue Nebs! Merging now.

Issue #565, opened 2022-08-24
Add sankey diagram to period analytics


kristoferlund:
Similar to the one found in the RAD analytics.
Givers to the left, receivers to the right.

Chart.js module here: https://github.com/kurkle/chartjs-chart-sankey

<img width="1056" alt="image" src="https://user-images.githubusercontent.com/9698363/186504714-97e9f8d9-19e4-44b7-8444-ee79ea0f4547.png">

kkatusic:
@kristoferlund I saw that from online example **priority** options left and I removed it. 
Now getting new error: _Uncaught RangeError: Maximum call stack size exceeded at nodeCount_

Bug reported: https://github.com/kurkle/chartjs-chart-sankey/issues/72

kkatusic:
Author fixed issue, @kristoferlund look a chart 😄 

![Praise](https://user-images.githubusercontent.com/1242145/190730500-b1ef0f1f-5cc9-4497-bfe0-b9ab73a39950.jpg)

Fix https://github.com/commons-stack/praise/tree/add/sankey-diagram

kristoferlund:
😂 a bit messy. hmm.. what to do?

kkatusic:
Restrict iviangita to praise someone again 😄

kkatusic:
Can you check on some other account this sankey diagram?

kristoferlund:
I 6 months have a look and see if component has been upgraded to better serve our needs.

PR #567, opened 2022-08-27
WIP: Sankey Chart for period givers/receivers

kristoferlund:
Chart of the praise flow between Givers and Receivers.

Resolves: #565


kristoferlund:
@kkatusic, like you said, there does not seem to be any options that would make this diagram look more like the old python one. Perhaps we should just skip this diagram? It is not the most important anyway.

kristoferlund:
Closing this, component doesn't serve our needs.

PR #573, opened 2022-09-08
Fix/evenly assign quantifiers prevent overlap

### LINKED ISSUE #535: Quantify screen lists wrong number of praise
There is a mismatch between the number in the top section and the number of praise in the actual list. Which is right? Perhaps wait and see if this is resolved with next version after the big Recoil refactor.

@usama9500 posted this description:

> Hi @kristofer, I was quanitifying the praise. I am not sure its a bug or something. For two of the people in my dashboard, Number of praise items shown are 2 but there is only one praise item shown for quantification.  You can check this screenshot:

image

image

mattyg:
Resolves #535

- Verify that no quantifier has been assigned to same receiver multiple times, before generating assignments
- Refactor `prepareAssignmentsEvenly` to prevent a case where a quantifier could be assigned to the same receiver twice
- Improve doc comments explaining `prepareAssignmentsEvenly`

**Post-Mortem**
Previously in `prepareAssignmentsEvenly` we:

1. Generated a list of unique groups of receivers, duplicating each receiver for redundant assignments
2. Attempted to evenly distribute the groups of receivers among all quantifiers, based on the total number of praise in the group of receivers (i.e. a quantifier could be assigned to multiple groups of receivers).

This allowed a case where a quantifier could be assigned a duplicate receivers, because multiple groups of unique receivers could potential contain a duplicate between them.

With this refactor, we swapped the order of these steps. Now we:
1. Attempt to evenly distribute receivers among all quantifiers, based on the total amount of praise each receiver has (i.e. a quantifier could be assigned multiple receivers).
2. Duplicating the assignments, rotating the order, for each redundant quantification.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -10,5 +10,15 @@ docker exec -i mongodb-praise /usr/bin/mongodump --authenticationDatabase admin 
 
 **Restore**
 ```
-docker exec -i mongodb-praise sh -c 'mongorestore --authenticationDatabase admin -u DB_ROOT_USER -p DB_ROOT_PASSWORD --db praise_db --archive' < database-backup-BACKUP_DATE.archive
+docker exec -i mongodb-praise sh -c "mongorestore --authenticationDatabase admin -u DB_ROOT_USER -p DB_ROOT_PASSWORD --db praise_db --archive" < database-backup-BACKUP_DATE.archive
+```
+
+### Connect to mongodb database running on docker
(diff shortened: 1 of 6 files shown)


kristoferlund:
Great work @mattyg!! I have tested it as best I can with some different period/quantifier settings. Assignments look flawless! 

I added a couple of small frontend additions to better handle error messages when quantifiers cannot be assigned. 

Merging!

PR #591, opened 2022-09-20
Simplify the setup script, capture only what is absolutely needed #576

### LINKED ISSUE #576: Simplify the setup script, capture only what is absolutely needed

nebs-dev:
Resolves #576

- Simplified setup script
- Moved all env variables to one file in the root folder


packages/setup/src/index.ts
```diff
@@ -165,65 +105,39 @@ const run = async (): Promise<void> => {
...
-    MONGO_PASSWORD: answers.MONGO_PASSWORD,
-  };
-  await setupAndWriteEnv(
-    '/usr/praise/.env.template',
-    '/usr/praise/.env',
-    rootEnv
-  );
-
-  const apiEnv = {
+    MONGO_INITDB_ROOT_USERNAME: randomString(12),
```
kristoferlund, line 118: Don't generate random usernames for root username and username. Use the ones specified in `.env.template`

packages/setup/src/index.ts
```diff
@@ -165,65 +105,39 @@ const run = async (): Promise<void> => {
...
+    MONGO_INITDB_ROOT_PASSWORD: randomString(12),
```
kristoferlund, line 119: Passwords should not be regenerated on each run

packages/setup/src/index.ts
```diff
@@ -1,41 +1,17 @@
...
-});
-
-// Frontend
-dotenv.config({
-  path: '/usr/praise/packages/frontend/.env.template',
-  override: true,
-});
-dotenv.config({ path: '/usr/praise/packages/frontend/.env', override: true });
+dotenv.config({ path: '/usr/praise/.env.template', override: false });
+dotenv.config({ path: '/usr/praise/.env', override: false });
```
kristoferlund, line 14: Override should be `true`, right? The template should override any pre-existing env settings. And the "real" `.env` should override the template.

kristoferlund (changes requested):
Migration from current setup needs to be handled:
- Check if there are `.env` files in the package directories
- If there are, merge those with the main `.env`, then delete them
- Then load main `.env`
- Start setup

PR #592, opened 2022-09-21
Adding class to img tag

kkatusic:
Adding custom css classes to image tag if it exist inside praise string/message


kristoferlund (approved):
Nice simple fix! Merging.

PR #613, opened 2022-10-02
Add/setting default values

### LINKED ISSUE #580: Admin should be able to reset settings to default values in the praise dashboard settings interface
- On hover on the title of a setting, an icon button appears that allows an admin to reset a setting to the default value: image
- Button looks and works similar to the quantifier replacement button: image
- Add `defaultValue` to API interface `Setting` and `SettingDto`.
- Populate all default values using a database migration
- Use the `/settings/:id/set` to reset value, or create new endpoint `/settings/:id/reset`

kkatusic:
Closes #580 

@kristoferlund I need review. I didn't attached some fields and checkboxes. thx


kristoferlund (changes requested):
Nice! I made some small change to default values and texts. 

One thing I would like to adjust. Let's make the reset buttons invisible until you hover the setting name. And also do mouseover effect similar to how it is done on periods / quantifiers page.

kristoferlund:
Nice! I made some small change to default values and texts. 

One thing I would like to adjust. Let's make the reset buttons invisible until you hover the setting name. And also do mouseover effect similar to how it is done on periods / quantifiers page.

kristoferlund (approved):
Looks good @kkatusic! Merging

PR #619, opened 2022-10-04
Update to discord.js v14

Vyvy-vi:
- This Pr bumps the discord.js version and fixes all the breaking changes
- There are also some minor refactors to make things cleaner

Added by Kristofer
- Most likely closes: #524

PR #621, opened 2022-10-05
Logging system rework

### LINKED ISSUE #584: Rework the logging system
- Make sure logging is running as intended
- Make sure logs are accessible 
- Document how to access logs
- Log rotation?

nebs-dev:
Resolves #584 

- removed jet-logger 
- installed https://github.com/bithavoc/express-winston ( paired with https://github.com/winstonjs/winston-daily-rotate-file it has option remove old log files out of box ). It has a lot of configuration options that we need 
- created morgan middleware that is using winston logger, so now every error that is thrown will be logged in into the console & file


packages/api/src/shared/logger.ts
```diff
@@ -0,0 +1,54 @@
...
+  warn: 1,
+  info: 2,
+  http: 3,
+  debug: 4,
+};
+
+const level = (): string => {
+  const env = process.env.NODE_ENV || 'development';
+  const isDevelopment = env === 'development';
+  return isDevelopment ? 'debug' : 'warn';
```
kristoferlund, line 15: Could we allow production log level to be specified in .ENV? For development it can always be set to debug.

packages/discord-bot/src/commands/activate.ts
```diff
@@ -1,5 +1,5 @@
 import { SlashCommandBuilder } from '@discordjs/builders';
-import logger from 'jet-logger';
+import { logger } from 'api/src/shared/logger';
```
kristoferlund, line 2: Imports should be made from `/api/dist`.

kristoferlund (changes requested):
Looks good, just requested two small changes.

kristoferlund:
Also @nebs-dev, please see merge conflicts since I merged #619.

packages/api/src/shared/logger.ts
nebs-dev: ok.. i didn't want to new LOGGING_LEVEL variable to be mandatory so i set up default level for the production also `warn`
It will log errors and warnings

kristoferlund (approved):
Great! Merging.

Issue #623, opened 2022-10-05
Add `buffer` polyfills for rainbowkit to work correctly


kristoferlund:
https://github.com/rainbow-me/rainbowkit/blob/main/examples/with-vite/src/polyfills.ts

kkatusic:
Fixed 
![buffer](https://user-images.githubusercontent.com/1242145/194390624-013f6e63-204c-4737-bcc7-9ea2a2fb0ace.jpg)

PR #624, opened 2022-10-06
Added buffer for WalletConnect login option

kkatusic:
This buffer is used for running QR code for connection other type of the wallets #623


kristoferlund (approved):
Nice! Merging.

PR #631, opened 2022-10-10
Organise settings interface

### LINKED ISSUE #586: Organise settings interface
- Allow for grouping and sorting of the settings interface
- Each section should have a header and possibly also a description
- Wireframes needed

nebs-dev:
Resolves #586 

- migration for the new settings field `section`
- modified types and transformer
- created separate components for each setting group

### DIFF
--- a/packages/frontend/src/pages/Settings/components/SettingsSection.tsx
+++ b/packages/frontend/src/pages/Settings/components/SettingsSection.tsx
@@ -1,0 +1,21 @@
+interface Props {
+  header: string;
+  children?: JSX.Element;
+  description?: string;
+}
+
+export const SettingsSection = ({
+  header,
+  description,
+  children,
+}: Props): JSX.Element => {
+  return (
+    <div>
+      <div className="mb-8">
+        <div className="mb-2 text-xl">{header}</div>
+        <div className="text-sm">{description}</div>
+      </div>
+      {children}
+    </div>
+  );
+};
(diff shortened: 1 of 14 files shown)


nebs-dev:
@kristoferlund can you take a look at this? 
I separated each setting group into components. I know components are almost the same but i though it is more clean this way and also maybe sometimes in future we would like to modify these screens more
Also i added DB field because maybe we will add option to move fields from one section into other and than we only need to change this new `section` value

packages/frontend/src/pages/Settings/components/SettingsSection.tsx
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export const SettingsSection = ({
+  header,
+  description,
+  children,
+}: Props): JSX.Element => {
+  return (
+    <div>
+      <div className="mb-8">
```
kristoferlund, line 14: Margin 5 is the "default" margin we use in most places.

packages/frontend/src/pages/Settings/components/SettingsSection.tsx
```diff
@@ -0,0 +1,21 @@
...
+        <div className="mb-2 text-xl">{header}</div>
```
kristoferlund, line 15: Use `<h2>`

kristoferlund (changes requested):
Nice!

I agree with breaking out each settings group into components, looks good! 

Made some small styling comments.

One thing I think you should change though. We use the term `group` and `SettingsGroup`. `section` is very closely related to the groups, I think `subgroup` would more clearly show the connection.

kristoferlund (approved):
Nice Nebs! Merging!

PR #634, opened 2022-10-17
wip: add api key feature

### LINKED ISSUE #572: Allow access to the API using an API key
Currently, to access the Praise API you need to authenticate by signing a message using Metamask or other ETH wallet. This is all well and good for individual users. But, we would like to open up API access to external systems as well. In this case Metamask authentication is not ideal.

We would like to add API key as an authentication method to allow external systems _read only_ (to begin with) access to the API.

## Settings 
- Generate and delete API key
- An API key can be given `full` or `limited` access.
  - `full` - includes everything, similar to what an ADMIN gets
  - `limited` - similar to what a USER gets 

## API access 
- API key is sent as a request header
- API key gives read only access to all applicable endpoints

saucecodee:
Resolves #572


packages/api/src/api-key/controllers.ts
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @param {TypedRequestQuery<QueryInputParsedQs>} req
+ * @param {TypedResponse<PaginatedResponseBody<ApiKeyDto[]>>} res
+ * @returns {Promise<void>}
+ */
+export const all = async (
+  req: TypedRequestQuery<QueryInputParsedQs>,
+  res: TypedResponse<ApiKeyDto[]>
+): Promise<void> => {
+  const apiKeys = await ApiKeyModel.paginate({
```
kristoferlund, line 28: We don't need to do pagination for api keys as we don't expect there to be a large number of keys

packages/api/src/api-key/controllers.ts
```diff
@@ -0,0 +1,70 @@
...
+  const apikey = await ApiKeyModel.findById(id);
+  if (!apikey) throw new NotFoundError('Api key');
+
+  const result: ApiKeyDto = {
+    _id: apikey._id,
+    access: apikey.access,
+    apikey: apikey.apikey,
+    name: apikey.name,
+    createdAt: String(apikey.createdAt),
+    updatedAt: String(apikey.updatedAt),
```
kristoferlund, line 56: Use ISOStgring format: `createdAt: createdAt.toISOString()`

kristoferlund (changes requested):
I like that you separate user roles and api key acess: ` authMiddleware(UserRole.USER, ApiKeyAccess.LIMITED)`.

Made two small comments.

PR #638, opened 2022-10-18
fix: remove "praise success" dm duplication when account not activated

### LINKED ISSUE #571: Praise receivers should not receive an activation message upon receiving praise
image

@Vyvy-vi, this was reported from Gnosis. The first message should not be sent to a praise receiver, right? One message seems enough and getting the first message with no context seems off.

Vyvy-vi:
This PR removes the `PRAISE_ACCOUNT_NOT_ACTIVATED_ERROR_DM` setting and omits to send separate messages to users. If they've been praised and haven't activated, the message sent to them has some text instructing them about how to activate.

fixes #571


kristoferlund (approved):
Nice! I made some small text changes. Merging.

PR #648, opened 2022-11-02
Add/log event module

### LINKED ISSUE #647: Implement the `logEvent` function in api2

saucecodee:
Fixes #647

- Create and abstract shared module
- Add `EventLog` module 
- Implement  `logEvent` function


packages/api2/src/event-log/event-log.controller.ts
```diff
@@ -0,0 +1,17 @@
+import { Controller, Get, Param } from '@nestjs/common';
+import { EventLogService } from './event-log.service';
+
+@Controller('event-log')
+export class EventLogController {
```
kristoferlund, line 5: The api enpoints are `/all` and `/types`.

packages/api2/src/event-log/entities/event-log.entity.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  @Prop({ type: String, required: true })
+  description: string;
+}
+
+// eventLogSchema.plugin(mongoosePagination);
+
+export const EventLogSchema = SchemaFactory.createForClass(EventLog);
+
+// export const EventLogModel = model<
```
kristoferlund, line 47: Pagination needs to be supported

packages/api2/src/_shared/shared.module.ts
```diff
@@ -0,0 +1,7 @@
+import { Module } from '@nestjs/common';
+import { SharedService } from './shared.service';
+
+@Module({
+  providers: [SharedService],
+})
+export class SharedModule {}
```
kristoferlund, line 7: What is the purpose of the shared module?

kristoferlund (changes requested):
Looks good so far, some things remain for it to be finished though.

packages/api2/src/_shared/shared.module.ts
saucecodee: Shared module contains utils, pipes etc that can’t be in a separate module but are needed in multiple modules

Issue #625, opened 2022-10-07
Switching tabs praise<>mypraise causes bottom scroll to trigger each time


kristoferlund:

kristoferlund:
No more endless scroll!

PR #672, opened 2022-11-10
User profile praise don't always display correctly, top praise not being top praise

### LINKED ISSUE #670: User profile praise don't always display correctly, top praise not being top praise

nebs-dev:
Resolves #670 #625


kristoferlund (changes requested):
- Replace next/back with "Load more"
- The selection of Latest/Top should be remembered

PR #691, opened 2022-11-22
Kristofer 221122

kristoferlund:
- Created and `@IsEthAddress` validator to use on requests
- Renamed auth interfaces and DTOs to better follow/set naming standard
- Got unit tests working, have a look at mocked services and stubs for classes that will simplify testing a lot
- Got e2e tests working, minimal first implementation now runs for `/auth/nonce`


packages/api2/src/shared/validators.shared.ts
```diff
@@ -0,0 +1,32 @@
...
+
+@ValidatorConstraint({ name: 'IsEthAddress' })
+@Injectable()
+export class IsEthAddressRule implements ValidatorConstraintInterface {
+  validate(value: string) {
+    return ethers.utils.isAddress(value);
+  }
+
+  defaultMessage() {
+    return `($value) is not a valid Ethereum address.`;
```
nebs-dev, line 18: do you have `value` variable here?

PR #692, opened 2022-11-22
API 2 Settings

nebs-dev:


kkatusic:
Great work guys and girls :)

packages/api2/src/settings/settings.service.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  private uploadDirectory =
+    process.env.NODE_ENV === 'production' ? '/usr/src/uploads/' : 'uploads/';
+
+  async findAll(): Promise<Settings[]> {
+    const settings = await this.settingsModel.find().lean();
+    return settings.map((setting) => new Settings(setting));
+  }
+
+  async findOneById(_id: Types.ObjectId): Promise<Settings> {
```
kristoferlund, line 33: `UserService` returns `User | null` instead of throwing error for these simple find methods. We should pick one method and use everywhere. I have no clear opinion as to what is best. In the case of `userService.findOneByEth` returning null felt ok since we expect that method to be called with unknown addresses. Not finding a user is no exception then. But.. for the sake of clarity all over, and knowing that for most other methods not finding the object *is* an exception, perhaps we should stick to throwing exceptions?

kristoferlund (approved):
Looks, good! Great work Nebs! I made one small comment and then some things could perhaps be moved around. I don't like where customExportTransformer is in general. But that is ok for now, that feature needs to be upgraded to a more fully featured reporting and distribution module. Let's merge it and add more tests etc in another PR.

kristoferlund:
Looks, good! Great work Nebs! I made one small comment and then some things could perhaps be moved around. I don't like where customExportTransformer is in general. But that is ok for now, that feature needs to be upgraded to a more fully featured reporting and distribution module. Let's merge it and add more tests etc in another PR.

ps. Changes outside of api2 are all from merging with main branch, right? I haven't looked at those.

nebs-dev:
> Looks, good! Great work Nebs! I made one small comment and then some things could perhaps be moved around. I don't like where customExportTransformer is in general. But that is ok for now, that feature needs to be upgraded to a more fully featured reporting and distribution module. Let's merge it and add more tests etc in another PR.
> 
> ps. Changes outside of api2 are all from merging with main branch, right? I haven't looked at those.

Yes, I merged main to this branch and than you did the same for the api2 I think. After that I merged api2 to this branch and had some conflicts in yarn.lock, ran `yarn install` again so this is why we have a lot of changes in yarn cache
But I am not sure why it is showing all the changes for the `api`, `frontend`, etc

packages/api2/src/settings/settings.service.ts
nebs-dev: Yeah, i know that we talked about this. I was trying both and like you said - I think maybe this is more clear because how should we decide what is the simple method and what is not? :)
I would stick with this but we can do it like it is done in userService if you like

PR #695, opened 2022-11-25
API2 - PeriodSettings

nebs-dev:


packages/api2/src/settings/schemas/settings.schema.ts
```diff
@@ -33,6 +30,7 @@ export class Settings {
   get valueRealized(): string | boolean | number | number[] | undefined {
     if (!this || !this.value) return undefined;
 
+    // TODO: This could be rewritten to be more easily readable
```
kristoferlund, line 33: Could be rewritten as:

```
if (this.type === 'Integer') return Number.parseInt(this.value);
if (this.type === 'Float') return parseFloat(this.value);
... 
```

packages/api2/src/periodsettings/periodsettings.service.ts
```diff
@@ -0,0 +1,110 @@
...
+export class PeriodSettingsService {
+  constructor(
+    @InjectModel(PeriodSettings.name)
+    private periodSettingsModel: Model<PeriodSettingsDocument>,
+    private periodModel: Model<PeriodDocument>,
+    private utils: UtilsProvider,
+  ) {}
+
+  async findAll(periodId: Types.ObjectId): Promise<PeriodSettings[]> {
+    const period = await this.periodModel.findById(periodId);
```
kristoferlund, line 26: I would like us to avoid using the model directly and instead use `periodService.findOneById()`.

packages/api2/src/periodsettings/periodsettings.service.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (period.status !== PeriodStatusType.OPEN)
+      throw new ServiceException(
+        'Period settings can only be changed when period status is OPEN.',
+      );
+
+    const periodSetting = await this.periodSettingsModel.findOne({
+      _id: settingId,
+      period: periodId,
+    });
+    if (!periodSetting) throw new ServiceException('PeriodSettings not found.');
```
kristoferlund, line 67: "PeriodSetting not found."

packages/api2/src/settings/schemas/settings.schema.ts
```diff
@@ -8,9 +8,6 @@ export type SettingsDocument = Settings & Document;
 
 @Schema({
   timestamps: true,
-  // toJSON: {
-  //   virtuals: true,
-  // },
 })
 export class Settings {
```
kristoferlund, line 12: It should be `Setting` and `PeriodSetting`right? Not plural.

kristoferlund (changes requested):

packages/api2/src/periodsettings/periodsettings.service.ts
```diff
@@ -0,0 +1,111 @@
...
+  constructor(
+    @InjectModel(PeriodSetting.name)
+    private periodSettingsModel: Model<PeriodSettingDocument>,
+    private periodsService: PeriodsService,
+    private utils: UtilsProvider,
+  ) {}
+
+  async findAll(periodId: Types.ObjectId): Promise<PeriodSetting[]> {
+    const period = await this.periodsService.findOneById(periodId);
+    if (!period) throw new ServiceException('Period not found.');
```
kristoferlund, line 28: Since periodsService throws when not finding period we don't need to throw again here.

packages/api2/src/periodsettings/periodsettings.service.ts
```diff
@@ -0,0 +1,111 @@
...
+      .lean();
+    return settings.map((setting) => new PeriodSetting(setting));
+  }
+
+  async findOneById(
+    settingId: Types.ObjectId,
+    periodId: Types.ObjectId,
+  ): Promise<PeriodSetting> {
+    const period = await this.periodsService.findOneById(periodId);
+    if (!period) throw new ServiceException('Period not found.');
```
kristoferlund, line 41: Same here.

packages/api2/src/periodsettings/periodsettings.service.ts
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  async setOne(
+    settingId: Types.ObjectId,
+    periodId: Types.ObjectId,
+    req: Request,
+    data: SetPeriodSettingDto,
+  ): Promise<PeriodSetting> {
+    const period = await this.periodsService.findOneById(periodId);
+    if (!period) throw new ServiceException('Period not found');
```
kristoferlund, line 58: And here :)

kristoferlund (changes requested):

packages/api/src/praise/routes.ts
```diff
@@ -10,11 +10,13 @@ praiseRouter.getAsync('/:id', controller.single);
 praiseRouter.patchAsync(
   '/:id/quantify',
   authMiddleware(UserRole.QUANTIFIER),
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
   controller.quantify
 );
 praiseRouter.patchAsync(
   '/quantify',
   authMiddleware(UserRole.QUANTIFIER),
+  // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
kristoferlund, line 19: Hm. This is should probably not be here, but I see an error is raised otherwise. Let's leave it there since the issue will be no longer after migration to api2.

kristoferlund (approved):
Nice, merging this now. Don't forget tests still remain..

PR #718, opened 2022-12-02
Feat/authorization

kristoferlund:


kristoferlund:
Merging to api2 even though it is half baked to clear blockers with paths and imports.

PR #719, opened 2022-12-02
API 2 Migrations

### LINKED ISSUE #717: Migrate migrations to api2

nebs-dev:
Resolves #717


kristoferlund (approved):
Nice! I'm merging this without looking too much at it because I need the database connection etc for the e2e tets and seeding the test database.

PR #726, opened 2022-12-06
API2 Migrations

nebs-dev:


packages/api2/src/database/migration.ts
```diff
@@ -1,20 +1,53 @@
...
+    const db = await connectDatabase('localhost');
+    logger.log('Connected to database');
+
+    const migrator = new Umzug({
+      migrations: { glob: 'src/database/migrations/*.ts' },
+      storage: new MongoDBStorage({
+        connection: db.connection,
+        collectionName: 'migrations',
+      }),
+      logger: console,
```
kristoferlund, line 30: Define a logger that outputs using Nest logger.

packages/api2/src/database/migration.ts
```diff
@@ -1,20 +1,53 @@
...
  * @returns {Umzug}
  */
-const setupMigrator = (connection: Connection): Umzug => {
-  const migrator = new Umzug({
-    migrations: { glob: 'src/database/migrations/*.ts' },
-    storage: new MongoDBStorage({ connection, collectionName: 'migrations' }),
-    logger: console,
-  });
+const runDatabaseMigrations = async (app: INestApplication): Promise<void> => {
+  const logger = new Logger(PermissionsGuard.name);
```
kristoferlund, line 18: Other name needed. Suggestion: "Migrations"

packages/api2/src/database/migrations/13_settings_require_discord_role.ts
```diff
@@ -1,5 +1,5 @@
-import { SettingGroup } from '@/settings/types';
-import { SettingsModel } from '../../settings/entities';
+import { SettingGroup } from '@/settings/interfaces/settings-group.interface';
```
kristoferlund, line 1: Absolute imports in migrations using `@` did not work for me when running `yarn workspace api2 start`. Did it work for you?

packages/api2/src/database/schemas/settings/23_settings.schema.ts
```diff
@@ -0,0 +1,32 @@
+import { SettingGroup } from '@/settings/interfaces/settings-group.interface';
```
kristoferlund, line 1: Really like being able to see the evolution of the schema like this!

packages/api2/src/database/seeder/users.seeder.ts
```diff
@@ -1,29 +1,29 @@
-import { UserRole } from '@/users/interfaces/user-role.interface';
-import { User } from '@/users/schemas/users.schema';
-import { UsersService } from '@/users/users.service';
-import { faker } from '@faker-js/faker';
-import { Injectable } from '@nestjs/common';
+// import { UserRole } from '@/users/interfaces/user-role.interface';
```
kristoferlund, line 1: This class is new and being used by current tests.

packages/api2/src/users/users.service.ts
```diff
@@ -146,4 +147,30 @@ export class UsersService {
...
   }
+
+  /**
+   * Generate username from user account name
+   * If username is already taken than create one with discriminator
+   *
+   * @param userAccount
+   * @returns {Promise<string>}
+   */
+  generateUserNameFromAccount = async (
```
kristoferlund, line 158: Should the service class include methods that do not modify data? Or should those methods be in a utility class? But in that case we would like to make the utility class a provider. UsersUtilsProvider, hrm. Doesn't feel a 100% either. What do you think?

packages/api2/src/database/migration.ts
```diff
@@ -1,20 +1,53 @@
...
 
-  return migrator;
+  try {
+    const db = await connectDatabase('localhost');
```
kristoferlund, line 21: It shouldn't be needed to specify 'localhost' here, right?

packages/api2/src/database/migration.ts
```diff
@@ -1,20 +1,53 @@
...
+        usersService: app.get(UsersService),
+        periodsService: app.get(PeriodsService),
+        settingsService: app.get(SettingsService),
+        periodSettingsService: app.get(PeriodSettingsService),
+        utilsProvider: app.get(UtilsProvider),
+      },
+    });
+    logger.log('Migrator created');
+
+    require('ts-node/register');
```
kristoferlund, line 42: Is this required for the migrations to support TypeScript?

kristoferlund (changes requested):
Looks good! Some minor tweaks needed only.

kristoferlund (changes requested):
Looks good! Some minor tweaks needed only.

packages/api2/src/database/migration.ts
nebs-dev: yeah, because if I don't do that, I am getting this error - `Cannot use import statement outside a module`

PR #754, opened 2022-12-22
Added docker-compose installation script and modified scripts

mhmdksh:


kristoferlund (approved):
Is this using v2 version of Docker Compose? I thought `docker-compose` meant using  the old version of compose?

mhmdksh:
> Is this using v2 version of Docker Compose? I thought `docker-compose` meant using the old version of compose?
> 
> But if I have misunderstood that, please go ahead and merge, looks good!

@kristoferlund Yeah after checking it out, apparently docker compose is meant to be an upgrade to the original docker-compose to Go.
The versions though are still the same, I thought just using the normal docker-compose now is better with multi-instance setup, as (docker compose) needs specific configuration for every user running it (Which is not so convenient).

Before I merge this, I'll be adding 1 more commit, that will enhance the .env setup to be compatible with multi instance in which it make every instance running on the same server have unique names/ports etc.. to container duplication

nebs-dev (dismissed):

kristoferlund (approved):

PR #761, opened 2022-12-28
Enhanced the multi-setup scripts

mhmdksh:
- Read and configure existing config .env values if they existed
- https for single setup, http for multisetup
- removed duplicate setup.sh file (it's now setup-old.sh)
- added the .env.template to the script folder (In case the setup-old.sh is used)


kristoferlund (approved):
Looks good!

PR #768, opened 2023-01-02
API2 Period - create, update, close, praise list

nebs-dev:


kristoferlund (approved):
Looks good! There is some stuff left to do in terms of documentation, will make separate issues. 

And, `/periods/:id/praise` we should remove completely and instead add a `period` query filter option to `/praise`. Or, we keep it but make the response paginated.

PR #783, opened 2023-01-13
Add migration and user old schema to delete tokens

CarlosQ96:
Remove accessToken and RefreshToken from old Api that was saving it in the db, to remove security issues.
We no longer save it in Db only return in headers.


CarlosQ96:
Looking at old schema is doesn't contain indexes.

kristoferlund (approved):
Looks good! Merging

PR #787, opened 2023-01-19
API2 Periods Tests

nebs-dev:


kristoferlund (changes requested):
Quantification not working:
Screenshot 2023-01-25 at 15 47 20

Quantification objects are missing createdAt and updatedAt:
image

PR #798, opened 2023-01-26
Add endpoints for json and csv exporting

### LINKED ISSUE #780: API2 - Export endpoints
We should add export endpoints that export praise data either as json or csv.

New GET endpoints:
```
/api/periods/export
/api/praise/export
/api/quantifications/export
/api/users/export
/api/useraccounts/export
```

Querystring filters:
- All endpoints: `format`- `json` or `csv`, defaults to `csv`
- `/api/praise/export`
  - `startDate`, `endDate`, `periodId` 
- `/api/quantifications/export`
  - `startDate`, `endDate`, `periodId`

CarlosQ96:
Added export endpoints for Quantifications, User, UserAccounts, Praise, Periods.

Filtering by Date, and tests

Added filter by periodId.

Resolves #780


packages/api2/test/quantification.e2e-spec.ts
```diff
@@ -0,0 +1,240 @@
...
+        app,
+        adminUserAccessToken,
+      ).expect(200);
+      expect(response.text).toBeDefined();
+      expect(response.text).toContain('_id');
+      expect(response.text).toContain(String(quantifications[0].praise));
+      expect(response.text).toContain('score');
+      expect(response.text).toContain('scoreRealized');
+      expect(response.text).toContain('dismissed');
+      // expect(response.text).toContain('duplicatePraise');
```
CarlosQ96, line 233: This column is excluded when parsing to CSV for some reason, but not sure what is the reason.

packages/api2/test/userAccount.e2e-spec.ts
```diff
@@ -0,0 +1,150 @@
...
+      ).expect(200);
+      expect(response.text).toBeDefined();
+      expect(response.text).toContain(String(userAccounts[0]._id));
+      expect(response.text).toContain('_id');
+      expect(response.text).toContain('user');
+      expect(response.text).toContain('accountId');
+      expect(response.text).toContain('name');
+      expect(response.text).toContain('avatarId');
+      expect(response.text).toContain('platform');
+      // expect(response.text).toContain('activateToken');
```
CarlosQ96, line 145: This column is also excluded when parsing.

CarlosQ96:
Added periodId filters !!!!

packages/api2/src/shared/dto/export-request-options.dto.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Types } from 'mongoose';
+
+export class ExportRequestOptions {
+  @ApiProperty({
+    default: 'csv',
+  })
+  @IsOptional()
+  @IsString()
+  @Type(() => String)
+  format: string;
```
kristoferlund, line 14: Use an enum to enforce allowed options

packages/api2/src/praise/praise.controller.ts
```diff
@@ -51,6 +54,38 @@ export class PraiseController {
...
+    @Res({ passthrough: true }) res: Response,
+  ): Promise<Praise[] | undefined> {
+    const praises = await this.praiseService.export(
+      options.format,
+      options.startDate,
+      options.endDate,
+      options.periodId
+    );
+
+    if (options.format === 'json') return praises as Praise[];
```
kristoferlund, line 80: No need to do the `Content-Disposition` etc for json exports?

packages/api2/src/praise/praise.service.ts
```diff
@@ -90,6 +91,63 @@ export class PraiseService {
...
+   * @returns {Promise<[Date, Date]>}
+   * @throws {ServiceException}
+   *
+   **/
+  async selectDateFilterRange(
+    periodId: string,
+    startDate: string,
+    endDate: string,
+  ): Promise<[Date, Date]> {
+    if (periodId && startDate && endDate) {
```
kristoferlund, line 132: Query should fail if periodId and either startDate or endDate is specified.

kristoferlund (changes requested):
Nice work!! I left two small comments only.

packages/api2/src/praise/praise.service.ts
CarlosQ96: I added the response object to the controller, if the content type is csv I use content disposition.

```
res.set({
      'Content-Type': 'text/csv',
      'Content-Disposition': 'attachment; filename="users.csv"',
    });
 ``` 

I wasn't 100% sure what is the best way to do it in nestJs, I found that way

File: packages/api2/src/praise/praise.controller.ts (hunk 4 above)
Comment (CarlosQ96) [reply packages/api2/src/praise/praise.controller.ts:80]: I added content disposition in the controllers, using the response object. It's the way I could make it work.

Comment (CarlosQ96) [reply packages/api2/src/praise/praise.controller.ts:80]: I added the response object to the controller, if the content type is csv I use content disposition.

```
res.set({
      'Content-Type': 'text/csv',
      'Content-Disposition': 'attachment; filename="users.csv"',
    });
res.send(data)
 ``` 

I wasn't 100% sure what is the best way to do it in nestJs, I found that way

packages/api2/src/praise/praise.service.ts
CarlosQ96: Thanks!

packages/api2/src/shared/dto/export-request-options.dto.ts
CarlosQ96: Added!

kristoferlund:
I added some stricter type checking of inputs and some other small fixes. Merging now.

PR #807, opened 2023-02-01
Api2 period paginate

### LINKED ISSUE #773: Paginate `/periods/:id/praise` endpoint
We should remove the endpoint completely and instead add a period query filter option to `/praise`. Or, we keep it but make the response paginated.

kkatusic:
fix #773


kristoferlund:
I will wait with merging this until we do the frontend changes.

kkatusic:
Sure, thx.

kristoferlund:
I had a look at this now. Fetching period praise for a quantifier does not work. I think we might need to do an aggregate query  for the query and perhaps do the pagination manually. I will not prioritise this right now as it might take some head scratching  to sort out.

PR #815, opened 2023-02-03
Fix/log page filter

### LINKED ISSUE #788: Log page - filter is not working

saucecodee:
Fixed #788


kristoferlund (approved):
Good fix! I removed the morgan logger for now as we already are using the built in console logger. Currently we don't do request logging though, that is something we should consider. I'll make a new issue for that.

PR #817, opened 2023-02-06
Quantifications issues

### LINKED ISSUE #809: Quantifying, pulling slider to 0 causes error

### LINKED ISSUE #810: Quantifying causes username to disappear on quant screen

nebs-dev:
Resolves #810 
Resolves #809


kristoferlund (changes requested):
What is the status of this PR? Do we need to make the change to `notDismissedQuantifications`? No business rule has changed, 0 still means that quantification is not completed.

packages/api2/src/praise/services/praise.service.ts
```diff
@@ -240,7 +240,7 @@ export class PraiseService {
         praise._id as Types.ObjectId
       ).toString()}"`;
     } else {
-      if (!score) {
+      if (!score && score !== 0) {
```
kristoferlund, line 243: This condition can never happen

packages/api2/src/quantifications/services/quantifications.service.ts
```diff
@@ -205,29 +205,22 @@ export class QuantificationsService {
...
-    });
-
-    // If no quantifications are completed the score is 0
-    if (completedQuantifications.length === 0) return 0;
+    if (quantifications.length === 0) return 0;
 
     // Calculate the score for each quantification
     const scores = await Promise.all(
-      completedQuantifications.map((q) => {
+      quantifications.map((q) => {
```
kristoferlund, line 212: Rule is still - If a quantifiacation has the score 0 it means it is not completed and should not be included in the average. On the other hand, if q quant is marked as dismissed, it is counted as zero and included in the average.

kristoferlund (changes requested):
Marking as duplicate does not work in the frontend. Not for single praise item or multiple.

kristoferlund (approved):
Nice work Nebs! Merging.

Issue #820, opened 2023-02-07
Typo in period defaults setting


kristoferlund:
Change: "How many redundant quantifications are assigned to each praise?"
To: "How many quantifiers are assigned to each praise?"

<img width="730" alt="image" src="https://user-images.githubusercontent.com/9698363/217266439-34830841-0b10-4a18-806f-2d41f88e3ab0.png">

Maniyuvan:
Hi, I can work on updating this.

Maniyuvan:
Okay, will do that

PR #822, opened 2023-02-07
fix: resolved the typo in db migration description text

maniyuvan:
Issue [#820](https://github.com/givepraise/praise/issues/820)

Resolved the typo by taking a branch from api2.

### DIFF
--- a/packages/api/src/database/migrations/01_settings_add_label_description.ts
+++ b/packages/api/src/database/migrations/01_settings_add_label_description.ts
@@ -15,7 +15,7 @@ const settings = [
     key: 'PRAISE_QUANTIFIERS_PER_PRAISE_RECEIVER',
     label: 'Quantifiers Per Praise',
     description:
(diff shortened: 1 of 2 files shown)


kristoferlund (changes requested):
A new migration has to be added. Old migrations will not be re-run once finished.

kristoferlund:
Yes, please make PR to merge with branch api2 as all development is ongoing there. Soon we will merge all back to main and continue development there.

kristoferlund (changes requested):
* Change to api branch can be removed completely, we are not updating api any longer.
* When doing database migrations you should not modify old migrations as they only run once. Migration 01 ran a long time ago and will not run again on existing systems. Instead, you should allways  create a new migration, in your case named `32_xxx`.

maniyuvan:
Okay, I'll close this PR and create a new one with new migration file for api2 alone as you said.

PR #827, opened 2023-02-09
fix: typo resolved in the api2 package label description

### LINKED ISSUE #820: Typo in period defaults setting

maniyuvan:
Resolves #820 

@kristoferlund  Updated "How many redundant quantifications are assigned to each praise? to: "How many quantifiers are assigned to each praise?" in the **api2** package.


maniyuvan:
@kristoferlund hope, this is what you expected

kristoferlund:
> @kristoferlund hope, this is what you expected

Unfortunately no. :) 

All migrations need to be there still, renaming 01 to 32 wont work. What happens if migration 02 need the settings to be there and you have deleted that migration? For someone setting up the system for the first time all migrations will run, in numerical order. The migration you need to write, number 31, should only change the label in question, nothing else.

maniyuvan:
So no need to remove the old file, instead I need to create a file with 31_xx ?

maniyuvan:
@kristoferlund  I've update the PR as you sugguested.

packages/api2/src/database/migrations/31_settings_add_label_description.ts
```diff
@@ -0,0 +1,124 @@
+import { SettingModel } from '../schemas/settings/00_settings.schema';
+
+const settings = [
```
kristoferlund, line 3: The migration should only affect `PRAISE_QUANTIFIERS_PER_PRAISE_RECEIVER`. Now you include a whole list of settings that are not needed. Remove everything but `PRAISE_QUANTIFIERS_PER_PRAISE_RECEIVER`.

kristoferlund (changes requested):

packages/api2/src/database/migrations/31_settings_add_label_description.ts
maniyuvan: Updated the PR, please check now

PR #835, opened 2023-02-10
Quantifiers Assignments Fix

### LINKED ISSUE #832: Quantifier assignment don't work as expected

nebs-dev:
Resolves #832


kristoferlund (changes requested):
So, this fix works. But it should not be needed and it is not solving the root cause. 

If `PRAISE_QUANTIFIERS_ASSIGN_EVENLY` was a boolean then your added check is unnecessary. But it turns out it is still a string which it shouldn't be.

image

kristoferlund (approved):
👍👍

We might also consider fixing `settingsService.settingValue` so that it returns correctly type values. I'll add that as a separate issue.

PR #844, opened 2023-02-14
POST /praise endpoint 

### LINKED ISSUE #824: /praise endpoint

nebs-dev:
Resolves #824


packages/api2/src/praise/services/praise.service.ts
```diff
@@ -434,4 +437,170 @@ export class PraiseService {
...
+      const praiseData = {
+        reason: reason,
+        reasonRaw: reasonRaw,
+        giver: giverAccount._id,
+        sourceId: sourceId,
+        sourceName: sourceName,
+        receiver: receiver._id,
+      };
+
+      const praiseItem = await this.praiseModel.create(praiseData);
```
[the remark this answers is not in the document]
kristoferlund: Bot should ensure there are no invalid receivers before making call.

packages/api2/src/praise/services/praise.service.ts
```diff
@@ -434,4 +437,170 @@ export class PraiseService {
...
+
+    const selfPraiseAllowed: boolean = JSON.parse(
+      (await this.settingsService.settingValue(
+        'SELF_PRAISE_ALLOWED',
+      )) as string,
+    );
+
+    let warnSelfPraise = false;
+    if (!selfPraiseAllowed && receiverIds.includes(giverAccount.accountId)) {
+      warnSelfPraise = true;
```
kristoferlund, line 483: Throw ServiceException instead. In general, services should not take responsibility over which messages are displayed to the end user. You are now returning DISCORD_BOT specific messages. What if some other bot uses the service? Or, some other service?

packages/api2/src/praise/dto/praise-create-input.dto.ts
```diff
@@ -0,0 +1,30 @@
+import { UserAccount } from '@/useraccounts/schemas/useraccounts.schema';
+import { ApiProperty } from '@nestjs/swagger';
+import { Type } from 'class-transformer';
+import { IsNotEmpty, IsString } from 'class-validator';
+
+export class PraiseCreateInputDto {
```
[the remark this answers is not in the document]
kristoferlund: Forwarder need to be included as well.

packages/api2/src/praise/services/praise.service.ts
```diff
@@ -434,4 +437,170 @@ export class PraiseService {
...
+
+  /**
+   * Creates praises with a given receiver and reason
+   *  and returns the created praises
+   *
+   * @param {PraiseCreateInputDto} data
+   * @returns {Promise<PraiseCreateResponseDto>}
+   * @throws {ServiceException}
+   */
+  createPraiseItem = async (
```
kristoferlund, line 449: Handling of forwarder info needs to be here as well so we can use same endpoint for `/forward`.

kristoferlund (changes requested):
Overall, looks good but need a bit of tightening up.

So in general, we should keep the responses from the services quite clean. We cannot assume that we know what kind of bot we are seeing so we cannot return discord specific messages. The service is only responsible for saving data to the database, the service should not be involved in choosing which user message to show and why. 

Somewhere in the future we would like this endpoint to return only one praise item with many receivers. Currently, instead, we create copies of the praise which is not ideal, but that is how it is. So in our case, I think we should return an array of all those copies, right? That is the only way to find out the ids because we dont want to create any new data structures such as a list with IDs or anything. So let's just return an array with all the created praise items on success.

On failure, you just throw a service exception for the various failures:
 - self phrase is not allowed
 - invalid receiver.

```
201 OK
[
  {praise},
  {praise},
  {praise}
]

400 Bad Request
{error}
```

packages/api2/src/praise/services/praise.service.ts
nebs-dev: Ok, i think I fixed all other comments except this one.. it should be easy but I want to check with you first to be sure what needs to be done.

kristoferlund (approved):
Great @nebs-dev! I added some more tests and fixed some small issue. Now merging.

Issue #846, opened 2023-02-15
Fix api2 linting errors


kristoferlund:

mohammadranjbarz:
https://github.com/givepraise/praise/pull/843#issuecomment-1431203091

<img width="1440" alt="Screen Shot 1401-11-27 at 00 03 30" src="https://user-images.githubusercontent.com/9850545/219148048-15464f2d-6424-4f0b-a2ed-1192c680941d.png">

mohammadranjbarz:
We have some problems in running tests as well

<img width="1431" alt="Screen Shot 1401-11-30 at 10 03 01" src="https://user-images.githubusercontent.com/9850545/219932905-3598aace-6157-4251-a315-fdc399b77d65.png">

kristoferlund:
Yes, the unit testing coverage is quite low and not at all up to date. The e2e testing coverage is better. Fixing unit tests is not a big priority at the moment.

`yarn workspace api2 test:e2e`

mohammadranjbarz:
> Yes, the unit testing coverage is quite low and not at all up to date. The e2e testing coverage is better. Fixing unit tests is not a big priority at the moment.
> 
> `yarn workspace api2 test:e2e`
It has some failed test cases as well

<img width="1419" alt="Screen Shot 1401-12-01 at 09 20 57" src="https://user-images.githubusercontent.com/9850545/220020412-550b63ff-4c98-43f2-9652-7922659be836.png">

PR #863, opened 2023-02-20
Fix 846 linter and test errors

mohammadranjbarz:
related to #846

Screen Shot 1401-12-01 at 09 06 45
Screen Shot 1401-12-01 at 09 08 54


packages/api2/src/auth/strategies/eth-signature.strategy.ts
```diff
@@ -30,11 +30,8 @@ export class EthSignatureStrategy extends PassportStrategy(
    * @returns
    */
   async validate(identityEthAddress: string, signature: string): Promise<any> {
-    let user;
-    // Check if user exists
-    try {
-      user = await this.usersService.findOneByEth(identityEthAddress);
-    } catch (e) {
```
mohammadranjbarz, line 37: @kristoferlund This code has logical problem and it had caused failing some unit tests

kristoferlund (approved):
Looks good! I added some small things - removing all linting warnings etc. Merging.

PR #866, opened 2023-02-20
Maniyuvan/fix_api2_typo

kristoferlund:


kristoferlund:
@maniyuvan I had to create a new PR to add my changes. Now merging this, thanks for contributing!

Changes I made before merge:
- Changed db schema to a later version, schema `00_` is out of date.
- Changed `down()` method. No need need to do anything on downgrade since this is a textchange only
- Changed filename to use a number that was not already in use.

PR #868, opened 2023-02-22
Fix/period end edgcases

### LINKED ISSUE #818: You should not be able to start quantification for a period that has not ended
- UI should disallow it, button should not show
- API should check

saucecodee:
Fix #818


kristoferlund (approved):
@saucecodee I fixed a few things and am merging now:
- The comparisons now < periodEnd was going in the wrong direction
- Tests were missing and some test was failing because of new rules. Be sure to add tests when making changes and verify that all tests succeed before marking PR for review.

PR #885, opened 2023-03-02
Add api-types package

marthendalnunes:
This PR adds the `api-types` module, which generates a `schema.ts` file with the interfaces from api2's openapi spec. The`api-types` module is imported by the frontend and replaces the previously hardcoded dtos. While refactoring the frontend to accept the interfaces from the openapi spec, some inconsistencies in the swagger documentation were found and fixed.

A GitHub worklflow was added to build the `schema.ts` file on `api-types` on any change of the `openapi.json` file on `api2`, ensuring the api interfaces are up to date with the latest changes.


packages/frontend/src/model/activate/dto/activate-request-body.dto.ts
```diff
@@ -1,5 +1,3 @@
-export interface ActivateRequestBodyDto {
-  identityEthAddress: string;
-  accountId: string;
-  signature: string;
-}
+import { components } from 'api-types';
+
+export type ActivateInputDto = components['schemas']['ActivateInputDto'];
```
kristoferlund, line 3: Filenames should be aligned with type names, in this case `activate-input.dto.ts`.

packages/api-types/out/schema.ts
```diff
@@ -0,0 +1,1708 @@
...
+      /** @example 5f9f1b9b9b9b9b9b9b9b9b9b */
+      _id: string;
+      /** @example 0xAAB27b150451726EC7738aa1d0A94505c8729bd1 */
+      identityEthAddress: string;
+      /** @example 0xAAB27b150451726EC7738aa1d0A94505c8729bd1 */
+      rewardsEthAddress: string;
+      /** @example darth */
+      username: string;
+      /** @example ["USER"] */
+      roles: string;
```
kristoferlund, line 219: Hmm, this becomes string instead of string[] even though api2 schema looks correct

packages/frontend/src/model/auth/auth.ts
```diff
@@ -54,7 +54,8 @@ export const ActiveUserRoles = selector<string[]>({
   key: 'ActiveUserRoles',
   get: ({ get }): string[] => {
     const user = get(SingleUser(get(ActiveUserId)));
-    if (user) return user.roles;
+    const userRoles = (user?.roles as string[]) ?? [];
```
kristoferlund, line 57: Here, roles should be string[], right?

packages/frontend/src/pages/EventLogs/components/EventLog.tsx
```diff
@@ -79,16 +78,6 @@ export const EventLog = ({
             title={eventlog.type.description}
             className={eventLogTypeColors[eventlog.type.key]}
           />
-
-          {eventlog.hidden ? (
```
kristoferlund, line 83: Why this deleted?

kristoferlund (changes requested):
Looks great!! This cleanup was badly needed, thanks! I made some smaller comments. And, frontend doesn't build, there are some errors related to `user.roles`, user account ids etc

packages/frontend/src/pages/EventLogs/components/EventLog.tsx
marthendalnunes: @kristoferlund the frontend was expecting a `hidden` parameter from the `api/event-log`, but only the older version of the api returns this value, the api2 doesn't so I tought it was an legacy behaviour. We should either remove this part of the code in the frontend or add the property on api2

kristoferlund: Yes, it is kind of an edge case. We didn't want to show quant scores from other quantifiers during an ongoing quantification in the logs. But .. like you say .. the feature has not been brought over to the new api and I don't believe that to be a significant risk anymore anyway. Let's retire it like you did.

kristoferlund (approved):
Looks great, nice work! Merging.

Issue #884, opened 2023-03-01
Move error messages to a separated file


mohammadranjbarz:
We use lots of hard coded messages in our code especially when throwing exceptions, @kristoferlund  what do you think about do some refactor and move those messages to a separated file?

kristoferlund:
Yes 👍 

Then we should also add a code to every ServiceException type. Will make it easier in testing to check that right exception is thrown.

PR #897, opened 2023-03-13
Refactor error handling flow

mohammadranjbarz:
related #884

@kristoferlund 
Do you think adding something like 
```
export const errorMessages: { [key: string]: ErrorMessage } = {
  COMMUNITY_NOT_FOUND: {
    message: 'Community not found',
    httpStatusCode: 404,
    code: 1001,
  },
};

```
is a good way? although I don't know how we want to use `httpStatusCode` or `code` , I think message may can be enough, suggest first decide about what we really want to do and after that proceed refactoring

### DIFF
--- a/packages/api2/src/activate/activate.service.ts
+++ b/packages/api2/src/activate/activate.service.ts
@@ -7,6 +7,7 @@ import { UsersService } from '@/users/users.service';
 import { EventLogService } from '@/event-log/event-log.service';
 import { EventLogTypeKey } from '@/event-log/enums/event-log-type-key';
 import { User } from '@/users/schemas/users.schema';
+import { errorMessages } from '@/utils/errorMessages';
 
 @Injectable()
 export class ActivateService {
@@ -40,9 +41,7 @@ export class ActivateService {
     const { identityEthAddress, signature, accountId } = activateInputDto;
 
     if (!identityEthAddress || !signature || !accountId)
-      throw new ServiceException(
-        'identityEthAddress, signature, and accountId required',
-      );
+      throw new ServiceException(errorMessages.INVALID_DATA_FOR_ACTIVATION);
 
     const userAccountModel = this.userAccountsService.getModel();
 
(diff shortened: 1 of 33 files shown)


kristoferlund:
Oh, now saw this: 

```
throw new NotFoundException(errorMessages.communityNotFound.message);
```

In general I think we should avoid throwing HTTP exceptions from services. Service should not assume HTTP being the transport layer.  That is why we introduced the `ServiceException` as a transport agnostic exception.

mohammadranjbarz:
@kristoferlund  You are right, please see the changes again and give your feedback:
1. Now we define a data structure for our errors like 
```
export const errorMessages: { [key: string]: ErrorMessage } = {
  COMMUNITY_NOT_FOUND: {
    message: 'Community not found',
    httpStatusCode: 404,
    code: 1001,
  },
  COMMUNITY_IS_ALREADY_ACTIVE :{
    message: 'Community is already active',
    httpStatusCode: 400,
    code:1002,
  }
};

```
2. When want to throw error we pass an object from above array like this:
```
 throw new ServiceException(errorMessages.communityNotFound);
```

3. Modify `ServiceException` class to create our desired exception
```
export class ServiceException extends Error {
  httpsStatusCode: number;
  code: number;

  constructor(errorMessage: ErrorMessage ) {
    super(errorMessage.message);
    this.name = 'PraiseException';
    this.httpsStatusCode = errorMessage.httpStatusCode
    this.code = errorMessage.code
  }
}

```

4. At the end we create appropriate response based on our exception
```

@Catch(ServiceException)
export class ServiceExceptionFilter implements ExceptionFilter {
  catch(exception: ServiceException, host: ArgumentsHost) {
    const ctx = host.switchToHttp();
    const response = ctx.getResponse<Response>();
    const status = exception.httpsStatusCode || 400; // Bad Request

    response.status(status).json({
      statusCode: status,
      message: exception.message,
      code: exception.code,
      error: 'Bad Request',
    });
  }
}

```

kristoferlund:
Yes @mohammadranjbarz this looks great!

mohammadranjbarz:
@kristoferlund  the Action job keeps failing but I think it's not related to my changes

https://github.com/givepraise/praise/actions/runs/4462536917/jobs/7837193197
Screen Shot 1401-12-29 at 00 05 25

kristoferlund (approved):
Looks good! Merging.

PR #900, opened 2023-03-15
Discord bot switch to api2

### LINKED ISSUE #856: Discord Bot uses API instead of accessing database directly
- **Include support for multi community together with this issue**
- When bot command is called from Discord, `Guild ID` is included in the request
- Bot adds this information as a http header for `API` requests along with information about user agent.
    
```jsx
user-agent: DiscordBot/0.13.1
x-discord-guild-id: 0980987846534
x-api-key: jhoijjp9jpj09pojp9j
```

## Use cases

https://www.notion.so/giveth/Give-praise-using-Discord-Bot-2dc3710f4b644d47b9a1d6caba00cceb

Vyvy-vi:
This PR replaces all db interactions done directly by the bot with API calls made to the api2 package

resolves #856


packages/api/src/pre-start/env-required.ts
```diff
@@ -15,6 +15,7 @@ const requiredEnvVariables = [
   'API_URL',
   'DISCORD_TOKEN',
   'DISCORD_GUILD_ID',
+  'DISCORD_BOT_API_KEY',
```
kristoferlund, line 18: Use instead the more generic new env fields:

```
API_KEYS=
API_KEY_ROLES=
```

The setup script will generate these keys when run. Bear in mind that more than one key can be defned in those fields. Use the key with the role `API_KEY_DISCORD_BOT`

https://github.com/givepraise/praise/blob/7ad7e1ed6aac1cb55e3ad629843dceacd9154ae8/packages/setup/src/index.ts#L162-L163

packages/discord-bot/src/utils/getUserAccount.ts
```diff
@@ -1,27 +1,82 @@
...
+      platform: 'DISCORD',
+    },
+    {
+      headers: { 'x-discord-guild-id': guildId },
+    }
+  );
+  const response = await apiClient.get<UserAccount[]>(
+    `/useraccounts/?accountId=${user.id}`,
+    {
+      headers: { 'x-discord-guild-id': guildId },
```
kristoferlund, line 24: Header to send to api to connect call with the right community is `host`. 

See here for details: https://www.notion.so/giveth/Give-praise-using-Discord-Bot-2dc3710f4b644d47b9a1d6caba00cceb?pvs=4

packages/discord-bot/src/utils/settingsUtil.ts
```diff
@@ -0,0 +1,91 @@
...
+        process.env.API_PORT as string
+      }/api/settings/uploads/${value}`;
+    case 'JSON':
+      return value ? JSON.parse(value) : [];
+    default:
+      return value;
+  }
+}
+
+export const getDefaultSetting = async (
```
kristoferlund, line 32: Is this needed? All settings already include default values in the DB. 

https://github.com/givepraise/praise/blob/7ad7e1ed6aac1cb55e3ad629843dceacd9154ae8/packages/api/src/settings/schemas/settings.schema.ts#L71

Vyvy-vi: I added this as a fallback in case the API doesn't return a valid setting due to some reason.

kristoferlund (approved):
What a massive amount of work you have out in here @Vyvy-vi, thank you! I am merging this now and we can create separate issues for what remains to be done.

PR #901, opened 2023-03-15
Bugs/activation login bugs

### LINKED ISSUE #895: Fix activation / login bugs
We are currently getting 5 testing errors for api2. Errors are related to useraccounts and activation endpoints.

One that I have been unable to solve is this one:
`POST /api/useraccounts › 400 when invalid fields`

The api should not allow attributes that are not in the Dto. But, no error message when adding invalid field in this test.

Hint: This is related to settings in `main.ts`:
```
    new ValidationPipe({
      transform: true,
      whitelist: true,
      forbidNonWhitelisted: true,
      //forbidUnknownValues: true, //TODO: Enabling this causes an error in the validation pipe - Investigate!
    }),
```

nebs-dev:
Resolves #895


kristoferlund (approved):
Looks good! Merging

PR #917, opened 2023-03-28
Replace api with api2

mohammadranjbarz:


mohammadranjbarz:
Oh I didnt see it's merged now

PR #954, opened 2023-04-18
Testing mobile wallet connect

### LINKED ISSUE #942: Unable to login on mobile
Clicking on the Metamask button triggers no action at all.

kkatusic:
Fixes #942


kristoferlund:
@kristoferlund will ask @marthendalnunes about how to best use RainbowKit

.yarn/cache/@rainbow-me-rainbowkit-npm-0.10.0-b882e0cf5d-57dd737fb5.zip
marthendalnunes, line 1: @kkatusic add `./yarn` to `.gitignore`. Yarn cache files shouldn't be committed to the repo

packages/frontend/package.json
```diff
@@ -21,7 +21,7 @@
     "@headlessui/react": "^1.4.1",
     "@json2csv/plainjs": "^6.1.2",
     "@mui/material": "^5.3.0",
-    "@rainbow-me/rainbowkit": "^0.10.0",
+    "@rainbow-me/rainbowkit": "^0.12.8",
```
marthendalnunes, line 24: Good catch on upgrading rainbowkit to the latest version

packages/frontend/src/providers/Web3Provider.tsx
```diff
@@ -11,19 +11,22 @@ import { WagmiConfig, configureChains, createClient } from 'wagmi';
...
+import { alchemyProvider } from 'wagmi/providers/alchemy';
 
 const { chains, provider } = configureChains(
   [mainnet],
   [
+    alchemyProvider({ apiKey: 'F2zZplZomYztFDGt01AjNSXPtFfM3APh' }),
+    publicProvider(),
     infuraProvider({
       apiKey: '1f04715c0a10492483047083a5c3edd2',
     }),
```
marthendalnunes, line 23: Keep the public provider as the last option
```suggestion
    alchemyProvider({ apiKey: 'F2zZplZomYztFDGt01AjNSXPtFfM3APh' }),
    infuraProvider({
      apiKey: '1f04715c0a10492483047083a5c3edd2',
    }),
    publicProvider(),
```

marthendalnunes: @kkatusic NEVER EVER put API keys directly into the code. Use env variables here and add them to `.env.example` so other devs can know which env variables are necessary to run the project.

marthendalnunes: @kkatusic the project is very simple and using only Alchemy as a provider is ok, what was the reasoning behind also addind Infura?

packages/frontend/src/providers/Web3Provider.tsx
```diff
@@ -11,19 +11,22 @@ import { WagmiConfig, configureChains, createClient } from 'wagmi';
...
-    publicProvider(),
   ]
 );
 
 const { connectors } = getDefaultWallets({
   appName: 'Praise',
+  projectId: '969fbaacb59c1313def58b6181921bec',
```
marthendalnunes, line 29: No hardcodes, please. Use env vars

marthendalnunes (changes requested):

packages/frontend/src/providers/Web3Provider.tsx
marthendalnunes: @kkatusic sorry, I mismatched the name, I meant to ask why you added the alchemy provider if only infura was being used.

marthendalnunes: @kkatusic the scope of the change is to fix the mobile issue with rainbowkit, not a migration of providers, please make each PR address a specific problem. If you think that changing the provider is relevant here, please add a new PR with this scope and justify the change. It's ok to keep with the current provider, especially because that doesn't affect the issue we're discussing on this PR.

kristoferlund (approved):
Made a small change to the alchemy key naming. 
Merging. Thanks for looking into this @kkatusic and @marthendalnunes !!

PR #960, opened 2023-04-24
Chore: Remove tracked yarn cache files and add .yarn/cache to .gitignore

marthendalnunes:
This PR addresses the issue of tracked yarn cache files in the repository. To keep the repository size smaller and prevent unnecessary tracking of cache files, we are removing the tracked yarn cache files and adding .yarn/cache to the .gitignore file.

**Changes**
- Remove tracked yarn cache files from the repository
- Add `.yarn/cache` to the `.gitignore` file to prevent future tracking


kristoferlund (approved):
👍👍
Merging

Issue #968, opened 2023-04-25
New api endpoint `/api/community/isNameAvailable`


kristoferlund:
When editing the create community form it would be nice to get instant feedback on whether the name is available or not. This can be done by adding a new api endpoint that checks if a community name is available.

kristoferlund:
For you @mohammadranjbarz ?

mohammadranjbarz:
> For you @mohammadranjbarz ?

Ok, thanks

PR #969, opened 2023-04-26
Implement GET /api/communities/isNameAvailable

### LINKED ISSUE #967: Change returned errors for create community
We have two scenarios where the name you want to give a community is not allowed:
1. The name is already taken by another community
2. The name is a reserved word

Both these errors should return the same error message and code - "Community name is not available".

This does not mean that the error message for invalid community names should be changed.

mohammadranjbarz:
* Implement #968 
* I made the endpoint public so it has no restriction and security for that
* I changed ` 'jest-extended/prefer-to-be-false': 'off'` in `.eslintrc` because it was creating lots of issues when formatting codes
* Fix #967 through this PR


mohammadranjbarz:
@kristoferlund  Now tests passed and this PR is ready for review

Issue #1021, opened 2023-05-17
Period end should have same date format as Period start


kristoferlund:
Later we will try to use localized dates all over but for now let's stick with the pattern "YYYY-MM-DD".

![Image](https://user-images.githubusercontent.com/9698363/238960855-4cd68f49-9eac-4201-af44-67b7f0c3dc5d.png)

ujwalkumar1995:
Hey @kristoferlund 
I would like to work on this. Could you please assign?

kristoferlund:
Welcome @ujwalkumar1995! Let me know if you have any issues setting up Praise locally. 

Prerequisite: Docker

```
nvm use
yarn
yarn mongodb:start
yarn setup
yarn workspace api start:debug
yarn workspace frontend start
```

kkatusic:
![346289462_276155481513039_4362635974469208072_n](https://github.com/givepraise/praise/assets/1242145/9f8d52f1-19ff-47c3-b16f-10da478daeed)

kristoferlund:
Partially fixed in #1025 

Issue remains that display format of input field is wrong. Possible solutions:
- Use oter date input component
- Do some sort of frontend trick where what is displayed is not actually the date input field

@ujwalkumar1995 unassigning you from this unless you want to keep at it.

PR #1025, opened 2023-05-18
Fix date format for end date to match start date

ujwalkumar1995:
Hey @kristoferlund 
This PR is for the issue #1021 
Let me know in case any changes are required.

### DIFF
--- a/packages/frontend/src/pages/PeriodDetails/components/PeriodDetailsHead.tsx
+++ b/packages/frontend/src/pages/PeriodDetails/components/PeriodDetailsHead.tsx
@@ -54,7 +54,7 @@ export const PeriodDetailsHead = (): JSX.Element | null => {
       </div>
       {!isAdmin ? (
         <>
-          <div>Period end: {formatIsoDateUTC(period.endDate)}</div>
+          <div>Period end: {formatIsoDateUTC(period.endDate, DATE_FORMAT)}</div>
           <div>Number of praise: {period.numberOfPraise}</div>
         </>
       ) : (


kristoferlund (changes requested):
Thanks for the quick fix. The period end date is not shown only there though. Look a few lines below and you will see `<PeriodDateForm />` where date also is shown. Rule is like this: 

- If user is regular user, display date in div
- If user is admin, display a date input allowing the admin to change date

ujwalkumar1995:
Hey @kristoferlund 
Thank you for the feedback. I assume you want me to try and set a format on the input of the date.. I did some research and as per below docs it says 

> Note: The displayed date format will differ from the actual value — the displayed date is formatted based on the locale of the user's browser, but the parsed value is always formatted yyyy-mm-dd.

Also mentioning the link for your reference.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date

Could you please let me know if I am missing something here?

kristoferlund:
@ujwalkumar1995, the displayed date should be formatted yyyy-mm-dd, the parsed value we don't need to touch, but maybe that is not achievable with the default date input?

ujwalkumar1995:
Hi @kristoferlund 
Yes that is at least what I could understand from my research. The display date in input takes the format from browser depending on user's locale.

kristoferlund:
@ujwalkumar1995 Thank you for looking into this! I will be merging your PR. We would like to solve this issue eventually, using another date component or some small trick to achieve desired result with the default input. If you would like to take this on, feel free. In any case, thanks for your contribution so far!

kristoferlund (approved):
Looks good! Merging.

PR #1067, opened 2023-06-06
Fixed target=blank for link inside any Praise

### LINKED ISSUE #1016: Link should open in new tab
Link inside praise should be opened in new tab instead of the same tab. Add `target="_blank"` to it.

Image

kkatusic:
Fix #1016


kristoferlund (approved):
👍👍

PR #1093, opened 2023-06-22
Increase Praise discord text length

### LINKED ISSUE #1080: Increase praise text length to 500 characters
Related to user question:

_I have a question, we've been having some issues with the characters limit that the praise bot has. Is there a way to bring it up to 500 characters?

People have been writing praise a little bit longer than 280 (which is the current limit) and when they send it the bot fails to record it and the text that they wrote gets lost
I can't seem to find a way to change this in the dashboard settings_

kkatusic:
Resolve #1080

### DIFF
--- a/packages/api/src/praise/schemas/praise.schema.ts
+++ b/packages/api/src/praise/schemas/praise.schema.ts
@@ -53,20 +53,20 @@ export class Praise {
     example: 'for making edits in the welcome text',
     type: 'string',
   })
-  @Prop({ required: true, minlength: 5, maxlength: 280 })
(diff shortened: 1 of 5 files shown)


Vyvy-vi (changes requested):
the praise schema in the api also needs to updated - 
https://github.com/givepraise/praise/blob/0599091100771bc214e99629525856eba10ae8fe/packages/api/src/praise/schemas/praise.schema.ts#L56-L70

Vyvy-vi (approved):
Looks Good To Me!

Issue #1047, opened 2023-05-30
Send welcome email when creating a Praise community


kristoferlund:
- [ ] Research email sender platform
- [ ] Create API email sender service

kristoferlund:
Sendgrid

kristoferlund:
When interacting with the service you should be able to submit a template and some data.

- Template is a template key or a filename 
- Data is merged with template, for example `{user}` is replaced with `Mohammad`

Perhaps like this?

```
template = "welcome-email.md"

data = {
  username: Mohammad
}

emailService.sendEmail(template, data);

```

@mohammadranjbarz sounds ok?

mohammadranjbarz:
> When interacting with the service you should be able to submit a template and some data.
> 
> * Template is a template key or a filename
> * Data is merged with template, for example `{user}` is replaced with `Mohammad`
> 
> Perhaps like this?
> 
> ```
> template = "welcome-email.md"
> 
> data = {
>   username: "Mohammad"
> }
> 
> emailService.sendEmail(template, data);
> ```
> 
> @mohammadranjbarz sounds ok?

Yeah seems good, will look at it

PR #1099, opened 2023-06-29
WIP Write basics of email service

mohammadranjbarz:
related to #1047

@kristoferlund I couldnt test it because we need to verify our domain https://docs.sendgrid.com/for-developers/sending-email/sender-identity

And also I think it's better to manage templates ourselves with creating some HTMLs it can be easier to get to know their services and temlates


kristoferlund:
> @kristoferlund I couldnt test it because we need to verify our domain https://docs.sendgrid.com/for-developers/sending-email/sender-identity

@mohammadranjbarz I believe we already have verified domain. 

Screenshot 2023-06-30 at 13 57 40

> And also I think it's better to manage templates ourselves with creating some HTMLs it can be easier to get to know their services and temlates

Could you tell me more? What are the benefits of managing the templates in the code? Or perhaps, what are the drawbacks of using SendGrid templates? The big benefit is that we can give Justina or Ahmad or a designer the access to sendgrid and allow them to modify templates without us having to rebuild the code.

Final comment. We should follow the Nest.js design patterns and build the email sender service as a real service. Now you have constructed it as a utility function.

mohammadranjbarz:
> > @kristoferlund I couldnt test it because we need to verify our domain https://docs.sendgrid.com/for-developers/sending-email/sender-identity
> 
> @mohammadranjbarz I believe we already have verified domain.
> 
> Screenshot 2023-06-30 at 13 57 40
> > And also I think it's better to manage templates ourselves with creating some HTMLs it can be easier to get to know their services and temlates
> 
> Could you tell me more? What are the benefits of managing the templates in the code? Or perhaps, what are the drawbacks of using SendGrid templates? The big benefit is that we can give Justina or Ahmad or a designer the access to sendgrid and allow them to modify templates without us having to rebuild the code.
> 
> Final comment. We should follow the Nest.js design patterns and build the email sender service as a real service. Now you have constructed it as a utility function.

Yeah I agree with you, thanks for your feedbacks, I'll work on them

mohammadranjbarz:
@kristoferlund  I made the changes that you needed, but still I couldn't test it, will message you on DM about it

PR #1106, opened 2023-07-10
fix: hide activation link

Vyvy-vi:
This PR edits the activation flow the user is shown when praising for the first time. This makes the activation link ephemeral, so that others can't activate on that user's account.


kristoferlund (approved):
Great! Merging

Issue #1007, opened 2023-05-11
Implement refresh JWT token handling in frontend


kristoferlund:
Currently we redirect the user to a generic error handling page for most unexpected errors. The expiry of JWT tokens is an expected error and needs custom handling. 

When we get such an error we should delete the current token and redirect user to login page. 

ps. Later we will implement refreshToken handling to avoid this alltogether.

kristoferlund:
This is what you currently get. No good. :)

![Image](https://github.com/givepraise/praise/assets/9698363/179085ef-e347-4434-a8b2-e1a98f23ebe7)

kristoferlund:
Yes, all errors have unique codes. Should be ok to add that condition.

<img width="538" alt="image" src="https://github.com/givepraise/praise/assets/9698363/6ccf5d42-0ed2-49b4-824a-9b4e0be07de6">

kkatusic:
@kristoferlund I can reproduce error, but there isn't redirection to error page, 404, anymore.
Should we close this?

kristoferlund:
@kkatusic renamed this issue

PR #1120, opened 2023-07-24
Refresh accessToken when JWT access token has expired

mohammadranjbarz:
related #1007

### DIFF
--- a/packages/frontend/src/utils/axios.ts
+++ b/packages/frontend/src/utils/axios.ts
@@ -32,20 +32,23 @@ export const handleErrors = async (
 
   if (
     refreshToken &&
+    recoilPersist &&
     err?.response?.status === 401 &&
-
     // 1092, 1107 are the error codes for invalid jwt token that defined in backend
     (err?.response?.data?.code === 1092 || err?.response?.data?.code === 1107)
   ) {
     // delete the old token from localStorage to prevent infinite loop
-    delete recoilPersist!['ActiveTokenSet'];
+    delete recoilPersist['ActiveTokenSet'];
     localStorage.setItem(JSON.stringify(recoilPersist), 'recoil-persist');
     try {
       await requestApiRefreshToken({ refreshToken });
       toast('Please try again');
       return err;
-    } catch (error: any) {
-      console.log('refresh accessToken error', error?.response?.data);
+    } catch (error) {
+      console.log(
+        'refresh accessToken error',
+        (error as AxiosError)?.response?.data
+      );
     }
   }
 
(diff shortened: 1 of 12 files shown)


packages/frontend/src/utils/axios.ts
```diff
@@ -9,10 +10,45 @@ const isJsonBlob = (data): data is Blob =>
...
+  if (
+    refreshToken &&
+    err?.response?.status === 401 &&
+
+    // 1092, 1107 are the error codes for invalid jwt token that defined in backend
+    (err?.response?.data?.code === 1092 || err?.response?.data?.code === 1107)
+  ) {
+    // delete the old token from localStorage to prevent infinite loop
+    delete recoilPersist!['ActiveTokenSet'];
+    localStorage.setItem(JSON.stringify(recoilPersist), 'recoil-persist');
```
mohammadranjbarz, line 42: @kristoferlund 
It seems we just can use `recoil` inside components so I had to manually put/get to local storage

packages/frontend/src/utils/axios.ts
```diff
@@ -9,10 +10,48 @@ const isJsonBlob = (data): data is Blob =>
...
+  if (recoilPersist) {
+    refreshToken = JSON.parse(recoilPersist)?.ActiveTokenSet?.refreshToken;
+  }
+
+  if (
+    refreshToken &&
+    recoilPersist &&
+    err?.response?.status === 401 &&
+    // 1092, 1107 are the error codes for invalid jwt token that defined in backend
+    (err?.response?.data?.code === 1092 || err?.response?.data?.code === 1107)
```
kristoferlund, line 38: 1092 means unauthenticated user trying to access authenticated resource. This should not trigger a refresh.

packages/frontend/src/utils/axios.ts
```diff
@@ -9,10 +10,48 @@ const isJsonBlob = (data): data is Blob =>
...
-): AxiosError => {
+): Promise<
+  AxiosError<{
+    code: number;
+    message: string;
+    statusCode: number;
+  }>
+> => {
+  let refreshToken;
+  const recoilPersist = localStorage.getItem('recoil-persist');
```
kristoferlund, line 28: Use `getRecoil(ActiveTokenSet)`, see `auth.ts` for example.

kristoferlund (changes requested):
Sorry for slow review @mohammadranjbarz! 

I'd like to see two small changes, other than that all good!

Or btw. Have you tried how the user experience is when token expires and gets refreshed? Since the original request fails, the UI is left in an unclear state. Should we force a UI reload to make sure all looks good?

kristoferlund:
@mohammadranjbarz I finished this as I want to get it out to communities soon.

kristoferlund (approved):

mohammadranjbarz:
> @mohammadranjbarz I finished this as I want to get it out to communities soon.

Thanks @kristoferlund , sorry I was busy last week , I couldn't spend time on this

PR #1125, opened 2023-08-02
fix(temp): block praise in pvt threads

Vyvy-vi:
- This temporarily blocks the usage of praise in private threads (returns an appropriate error message for the user)
- An actual fix that makes praise work in private threads may take a little while.


kristoferlund (approved):
Great! Also, we need to think about this. Should Praise be available in private threads? It seems like a possible privacy issue. Praising someone in a private thread might lead someone to believe that praise stays "private" when in reality it will be posted to the common feed.

PR #1128, opened 2023-08-07
WIP: Transform userpage Urls 

### LINKED ISSUE #839: Transform user page urls
We would like to give all users more shareable user page links. 

Instead of this: 

`https://praise.tecommons.org/users/61f71fea102583655ce2fdf0`

We want to do this:

`https://praise.tecommons.org/[username]`

Vyvy-vi:
fixes #839


kristoferlund (approved):
Looks great @Vyvy-vi, merging this!

One note. There is nothing stopping a user from choosing a "colliding" username. `periods`or `reports` for instance. We should have a blocklist that prevents users from choosing these names. Let's add a separate issue and add to API.

PR #1134, opened 2023-08-17
Fixing short discord names

### LINKED ISSUE #1132: Unable to praise to users with short usernames
Discord allows for usernames with 3 characters
Praise requires 4

Let's align with Discord requirements: 2-32 characters

kkatusic:
Fix #1132

@kristoferlund check if everything ok. I checked only on two places is required for username length, user schema and user_account schema.

There isn't any test required by name length, on front and bot side.

Only that is somehow suspicion to me is this bot code, cc @Vyvy-vi 

https://github.com/givepraise/praise/blob/main/packages/discord-bot/src/utils/dmTargets.ts#L105

I don't know why we have here 24 fixed number.


kristoferlund:
Yes, that looks a bit suspicious :) 

Let's unify username and useraccount.names lengths to 4-32 characters. 

@kkatusic You need to look at users.service as well:

```
generateUserNameFromAccount
generateValidUsername
```

kristoferlund:
@kkatusic Oh yes, forgot. We use eth address as username in some cases, when user logs in to dashboard without first activating their account. So let's not change username maxlength.

kkatusic:
@kristoferlund than this is done?

Vyvy-vi:
Would we also need to account for the twitter length limit? (2-15)

kristoferlund (approved):
All looks good, tests passing. I added 2-32 requirements to dashboard as well.

